PR #14, opened 2021-03-17
Renames private to secret where possible

George-cl:
To align with our naming convention elsewhere we need to replace references to 'private' keys with 'secret'.
Two of the dependencies (elliptic and key-encoder) used call them private keys however, this is internal to the methods in the file - all client facing methods have been renamed to remove private key.


hoffmannjan:
@George-cl this can be closed ?

George-cl:
> @George-cl this can be closed ? 

Yeah go for it, we can do this later

PR #15, opened 2021-03-19
ECO-990 / CLValue2 refactor

hoffmannjan:
* missing `Result`/ `Option` type (replacing custom implementation by [ts-result](https://github.com/vultix/ts-results))
* still need to add more tests & tsdoc comments
* some parts are marked `TBD`


casper-client-sdk.js
```diff
@@ -0,0 +1,2 @@
...
...[p];d&&!f&&(y[s]=c)}}return c}},1405:(t,e,r)=>{"use strict";var n="undefined"!=typeof Symbol&&Symbol,i=r(5419);t.exports=function(){return"function"==typeof n&&"function"==typeof Symbol&&"symbol"==typeof n("foo")&&"symbol"==typeof Symbol("bar")&&i()}},5419:t=>{"use strict";t.exports=function(){if("function"!=typeof Symbol||"function"!=typeof Object.getOwnPropertySymbols)return!1;if("symbol"==typeof Symbol.iterator)return!0;var t={},e=Symbol("test"),r=Object(e);if("string"==typeof e)return!1;if("[object Symbol]"!==Object.prototype.toString.call(e))return!1;if("[object Symbol]"!==Object.prototype.toString.call(r))return!1;for(e in t[e]=42,t)return!1;if("function"==typeof Object.keys&&0!==Object.keys(t).length)return!1;if("function"==typeof Object.getOwnPropertyNames&&0!==Object.getOwnPropertyNames(t).length)return!1;var n=Object.getOwnPropertySymbols(t);if(1!==n.length||n[0]!==e)return!1;if(!Object.prototype.propertyIsEnumerable.call(t,e))return!1;if("function"==typeof Object.getOwnPropertyDescriptor){var i=Object.getOwnPropertyDescriptor(t,e);if(42!==i.value||!0!==i.enumerable)return!1}return!0}},7642:(t,e,r)=>{"use strict";var n=r(8612);t.exports=n.call(Function.call,Object.prototy
```
zie1ony: This file should be build by CI only, right?

src/lib/CLValue2/Bool.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from 'chai';
+import { Bool } from './Bool';
+
+describe('CLValue Bool implementation', () => {
+  it('Should be able to return proper value by calling .value()', () => {
+    const myBool = new Bool(false);
```
zie1ony: Test also `true`.

src/lib/CLValue2/List.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  it('Should throw an error when list is not correct by construction', () => {
+    const badFn = () => new List([1, 2, 3]);
+
+    expect(badFn).to.throw("Invalid data type(s) provided.");
+  });
+
+  it('Should be able to return proper values by calling .value() on List', () => {
+    const myBool = new Bool(false);
+    const myList = new List<Bool>([myBool]);
```
zie1ony: `<Bool>` is not required here. In general we shouldn't be using it like this in tests, as it's not available in JS.

src/lib/CLValue2/List.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    expect(myList.value()).to.be.deep.eq([myBool]);
+  });
+
+  it('Should able to create empty List by providing type', () => {
+    const mList = new List(new BoolType());
+    const len = mList.size();
+
```
zie1ony, line 48: Add tests for `mList.get(100)` and `mList.pop()` on empty list.

src/lib/CLValue2/Map.ts
```diff
@@ -0,0 +1,103 @@
...
+    Array.isArray(v) &&
+    Array.isArray(v[0]) &&
+    v[0].length === 2 &&
+    !!v[0][0].clType &&
+    !!v[0][1].clType
+  );
+};
+
+export class MapValue<K extends CLValue, V extends CLValue> extends CLValue {
+  v: Array<[K, V]>;
```
zie1ony: I think it would be better to use [Map](https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Map) here and return Map in `value()`

src/lib/CLValue2/PublicKey.ts
```diff
@@ -0,0 +1,90 @@
...
+// TODO: Tidy up almost the same enum in 
+// { SignatureAlgorithm } '../Keys';
+export enum PublicKeyTag {
+  ED25519 = 1,
+  SECP256K1 = 2
+}
+
+export class PublicKeyType extends CLType {
+  toString(): string {
+    return 'URef';
```
zie1ony: Should be `return 'PublicKey'`

src/lib/CLValue2/String.ts
```diff
@@ -0,0 +1,33 @@
+import { CLType, CLValue } from "./Abstract"; 
+
+export class CLStringType extends CLType {
+  toString(): string {
+    return 'String';
+  }
+}
+
+export class CLString extends CLValue {
```
zie1ony, line 9: I like `CLString` naming concept. Maybe all types should follow that?

src/lib/CLValue2/Tuple.ts
```diff
@@ -0,0 +1,78 @@
...
+    inner: Array<CLType>;
+
+    constructor(inner: Array<CLType>) {
+      super();
+      this.inner = inner;
+    }
+
+    toString(): string {
+      const innerTypes = this.inner.map(e => e.toString()).join(", ");
+      return `Tuple (${innerTypes})`
```
zie1ony: Tuple name should contain it's `size`.

src/lib/CLValue2/URef.ts
```diff
@@ -0,0 +1,83 @@
...
+      this.accessRights.toString(8)
+    ].join('-');
+  }
+
+  clType(): CLType {
+    return new URefType();
+  }
+
+  value(): Uint8Array {
+    return this.data;
```
zie1ony: `value()` should also return `accessRights`, right?

src/lib/CLValue2/Unit.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class Unit extends CLValue {
+  data: [] = [];
+
+  clType(): CLType {
+    return new UnitType();
+  }
+
+  value(): [] {
+    return this.data;
```
zie1ony, line 17: I would return `undefined` here.

src/lib/CLValue2/String.ts
hoffmannjan: Yeap, seems to be better convention than `MapValue` for example

src/lib/CLValue2/Tuple.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    clType(): CLType {
+      return new TupleType(this.data.map(e => e.clType()));
+    }
+  }
+  return [Tuple, TupleType];
+};
+
+export const [Tuple1, Tuple1Type] = generateTupleClasses(1);
```
zie1ony, line 76: Change to `CLTuple`

src/lib/CLValue2/Unit.ts
```diff
@@ -0,0 +1,20 @@
+import { CLType, CLValue } from "./Abstract"; 
+
+export class UnitType extends CLType {
```
zie1ony, line 3: Change to `CLUnit`

zie1ony (approved):
Great job.

PR #17, opened 2021-03-22
ECO-991 / toBytes draft implementation

hoffmannjan:


src/lib/CLValue2/Abstract.ts
```diff
@@ -6,3 +8,7 @@ export abstract class CLValue {
   abstract clType(): CLType;
   abstract value(): any;
 }
+
+export abstract class ToBytes {
```
zie1ony, line 11: We should use our own for array of bytes. I would use Uint8Array everywhere.

src/lib/CLValue2/Bool.ts
```diff
@@ -1,12 +1,12 @@
...
 
 export class BoolType extends CLType {
   toString(): string {
     return 'Bool';
   }
 }
 
-export class Bool extends CLValue {
+export class Bool extends CLValue implements ToBytes {
   v: boolean;
```
zie1ony: Rename the variable to describe it a bit more, like "value".

src/lib/CLValue2/List.ts
```diff
@@ -12,7 +15,8 @@ export class ListType<T extends CLType> extends CLType {
   }
 }
 
-export class List<T extends CLValue> extends CLValue {
+export class List<T extends CLValue & ToBytes> extends CLValue
+  implements ToBytes {
   v: Array<T>;
```
zie1ony: Rename the variable to describe it a bit more, like "vector".

test/lib/byterepr.test.ts
```diff
@@ -43,112 +43,150 @@ describe(`numbers' toBytes`, () => {
   it('should be able to serialize/deserialize u8', () => {
```
zie1ony, line 43: This file lints doesn't have to be a part of this PR.

PR #18, opened 2021-03-22
rpc-client-js -> @open-rpc/client-js

hoffmannjan:
* After spinning `nctl` - seems like everything is working 
Screenshot 2021-03-22 at 21 03 35


hoffmannjan:
Bumped version so we can release after merge @zie1ony

zie1ony (approved):

PR #19, opened 2021-03-25
ECO-1012 CLValue toBytes on all CLValue Data Types

hoffmannjan:


src/lib/CLValue2/Result.ts
```diff
@@ -0,0 +1,85 @@
...
+export class CLResultType extends CLType {
+  toString(): string {
+    return 'Result';
+  }
+}
+
+export class CLResult<T extends CLValue & ToBytes, E extends CLErrorCodes>
+  extends CLValue
+  implements ToBytes {
+  public data: T | null;
```
zie1ony: Same as with `Option`. Why not to use `Result<T, E>` everywhere to store inner value and be constructor arg?
```
// When constructing, the CLResult should be used.
let r = CLResult.Ok(123);

// Then can be converted to Result.
let v = r.value();

// Now I'd modify it, but its not hand to put it back.
```

I'd rather do:

```
let r = new CLResult(Ok("asd"));
let v = r.value();
let r2 = new CLResult(v);
```

src/lib/CLValue2/Result.ts
```diff
@@ -0,0 +1,53 @@
...
+const RESULT_TAG_ERROR = 1;
+const RESULT_TAG_OK = 1;
+
+export class CLResultType extends CLType {
+  toString(): string {
+    return 'Result';
+  }
+}
+
+export class GenericResult<T, E> {
```
hoffmannjan, line 23: Not sure do we need this `GenericResult` but IMHO its handy to extends into CLResult as we still have possibility to create different internal Result types.

zie1ony (approved):

PR #21, opened 2021-03-30
ECO-1016 CLValue fromBytes on all CLValue Data Types

hoffmannjan:


src/lib/CLValue2/Abstract.ts
```diff
@@ -11,3 +15,19 @@ export abstract class CLValue {
 export abstract class ToBytes {
   toBytes: () => Uint8Array;
 }
+
+export abstract class FromBytes {
+  static fromBytes: (bytes: Uint8Array) => ResultAndRemainder<CLValue & ToBytes & FromBytes, CLErrorCodes>
```
zie1ony, line 20: I don't want the concept of remainder to be our public API. From the point of the end users, they just want to check if all the bytes can be deserialized to given CLValue. I would rename this to `consume` and make module private (if possible) and `fromBytes` should return `Result<CLValue, CLErrorCodes>` if something happens or remainder is not empty.

src/lib/CLValue2/AccountHash.ts
```diff
@@ -1,6 +1,17 @@
...
+  CLValue,
+  ToBytes,
+  CLErrorCodes,
+  ResultAndRemainder,
+  ACCOUNT_HASH_LENGTH,
+  resultHelper
+} from './index';
+import { Ok, Err } from 'ts-results';
 
 export class CLAccountHashType extends CLType {
```
zie1ony, line 12: Missing `FromBytes`.

src/lib/CLValue2/AccountHash.ts
```diff
@@ -1,6 +1,17 @@
-import { CLType, CLValue, ToBytes } from './Abstract';
```
zie1ony, line 1: AccountHash is not CLType https://docs.rs/casper-types/1.0.1/casper_types/enum.CLType.html.
But I would keep it for now. Let's try to talk to core devs about adding it, as it's good idea to me.

src/lib/CLValue2/constants.ts
```diff
@@ -0,0 +1,15 @@
+/// The length in bytes of a [`AccountHash`].
+export const ACCOUNT_HASH_LENGTH = 32;
```
zie1ony, line 2: That could live in `AccountHash.ts`.

src/lib/CLValue2/Key.ts
```diff
@@ -18,7 +25,7 @@ export class CLKeyType extends CLType {
 // TBD: Maybe the first should be CLByteArray insted?
```
zie1ony, line 25: I like the implementation, but just want to make a note, that Key type has grown in the last 2 months. https://docs.rs/casper-types/1.0.1/casper_types/enum.Key.html Let's keep it as is, but revisit soon in future, because Key might change even more.

src/lib/CLValue2/Key.ts
```diff
@@ -18,7 +25,7 @@ export class CLKeyType extends CLType {
...
 type CLKeyParameters = Uint8Array | CLURef | CLAccountHash;
 
-export class CLKey extends CLValue {
+export class CLKey extends CLValue implements ToBytes {
```
zie1ony, line 28: Missing `FromBytes`.

src/lib/CLValue2/Map.ts
```diff
@@ -35,6 +47,8 @@ const isValueConstructor = (
   );
 };
 
+type KeyVal = CLValue & ToBytes & FromBytes;
+
 export class CLMap<K extends CLValue & ToBytes, V extends CLValue & ToBytes>
   extends CLValue
   implements ToBytes {
```
zie1ony, line 54: Missing `FromBytes`.

src/lib/CLValue2/Option.ts
```diff
@@ -45,9 +65,12 @@ export class GenericOption <T> {
   }
 }
 
-export class CLOption extends GenericOption<CLValue & ToBytes> implements CLValue, ToBytes {
+export class CLOption extends GenericOption<CLValue & ToBytes> implements CLValue, ToBytes, FromBytes {
```
zie1ony, line 68: Shouldn't it be `CLValue & ToBytes & FromBytes`?

src/lib/CLValue2/Option.ts
```diff
@@ -1,15 +1,35 @@
...
+  ResultAndRemainder,
+  CLU8,
+  resultHelper
+} from './index';
 
 const OPTION_TAG_NONE = 0;
 const OPTION_TAG_SOME = 1;
 
-export class CLOptionType extends CLType {
+export class CLOptionType<T extends CLType | null> extends CLType {
```
zie1ony, line 18: Event if the the Option is None it needs to have a type, so you have to remove that `null`. Constructor for none, could accept CLType to mark that.

src/lib/CLValue2/PublicKey.ts
```diff
@@ -1,23 +1,33 @@
...
+  resultHelper,
+  ResultAndRemainder,
+  ToBytes,
+  FromBytes,
+} from './index';
 import { decodeBase16, encodeBase16 } from '../Conversions';
 import { byteHash } from '../Contracts';
 
 // TODO: Tidy up almost the same enum in 
 // { SignatureAlgorithm } '../Keys';
```
zie1ony, line 15: Add those length checks. ED25519 - 32bytes, SECP256K1 - 33bytes.

src/lib/CLValue2/Result.ts
```diff
@@ -1,34 +1,55 @@
 import { concat } from '@ethersproject/bytes';
```
zie1ony, line 1: Missing `FromBytes` implementation.

src/lib/CLValue2/String.ts
```diff
@@ -30,7 +44,23 @@ export class CLString extends CLValue implements ToBytes {
...
+  static fromBytes(
+    rawBytes: Uint8Array
+  ): ResultAndRemainder<CLString, CLErrorCodes> {
+    const { result: CLU32res, remainder: CLU32rem } = CLU32.fromBytes(rawBytes);
+    if (!CLU32res.ok) {
+      return resultHelper(Err(CLErrorCodes.EarlyEndOfStream));
+    }
+    const len = CLU32res.val.value().toNumber();
+    const val = fromBytesString(CLU32rem!.subarray(0, len));
+    // console.log(len);
```
zie1ony, line 60: Remove those logs.

src/lib/CLValue2/Tuple.ts
```diff
@@ -35,21 +44,63 @@ abstract class GenericTuple extends CLValue {
...
     return this.data;
   }
+
+  toBytes(): Uint8Array {
+    return concat(this.data.map(d => d.toBytes()));
+  }
+
+  static fromBytes(
+    rawBytes: Uint8Array,
+    type: typeof CLTuple1Type | typeof CLTuple1Type | typeof CLTuple3Type
```
zie1ony, line 61: Second type should be `CLTuple2Type`.

src/lib/CLValue2/Tuple.ts
```diff
@@ -35,21 +44,63 @@ abstract class GenericTuple extends CLValue {
...
+      return resultHelper(Ok(new CLTuple2(val)), rem);
+    }
+    if (val.length === 3) {
+      return resultHelper(Ok(new CLTuple3(val)), rem);
+    }
+    return resultHelper(Err(CLErrorCodes.Formatting));
+  }
 }
 
 // TBD: Maybe replace this with classic implementation, it will be easier to validate by compiler eg. new Tuple1<Bool>(), new Tuple2<Bool, I32>();
```
zie1ony, line 87: Yes we want values on creation.

src/lib/CLValue2/Tuple.ts
```diff
@@ -1,9 +1,18 @@
+import { Ok, Err } from "ts-results";
```
zie1ony, line 1: > // TBD: Maybe replace this with classic implementation, it will be easier to validate by compiler eg. new Tuple1<Bool>(), new Tuple2<Bool, I32>();

Yes. I like how Numbers are implemented.

zie1ony (changes requested):

hoffmannjan:
@zie1ony all of the changes are included in #32

PR #23, opened 2021-04-01
emergency fix for delegator info

hoffmannjan:


zie1ony (approved):
I have tested it with mainnet and works fine.

PR #42, opened 2021-04-28
Support Signer's disconnectFromSite

zie1ony:
### Added

- Support `disconnectFromSite` method from the Signer.


George-cl (approved):
LGTM, thanks!

PR #61, opened 2021-05-28
1.3.4 - Signer.getVersion returns running version of Signer

hoffmannjan:
## Added

- `Signer.getVersion` returns running version of Signer.


George-cl (approved):
LGTM!

PR #64, opened 2021-06-03
casper-client-sdk 1.4.0 - New Signer.Sign method

hoffmannjan:
## Changed

- `Signer.sign` now requires deploy in JSON format, `public-key hex` of a sender and `public-key hex` of a target.

### DIFF
--- a/src/@types/casperlabsSigner.d.ts
+++ b/src/@types/casperlabsSigner.d.ts
@@ -18,10 +18,10 @@ interface CasperLabsHelper {
    * Send Deploy in JSON format message to Signer plugin to sign.
    *
    * @param deploy deploy in JSON format
-   * @param sourcePublicKey base64 encoded public key used to sign the deploy
-   * @param targetPublicKey base64 encoded public key used to sign the deploy
+   * @param sourcePublicKeyHex public key in hex format with algorithm prefix. Used to sign the deploy
+   * @param targetPublicKeyHex public key in hex format with algorithm prefix. Used to display hex-formatted address on the UI
    */
-  sign: (deploy: JSON, sourcePublicKey: string, targetPublicKey: string) => Promise<JSON>;
+  sign: (deploy: any, sourcePublicKeyHex: string, targetPublicKeyHex: string) => Promise<JSON>;
   
   /*
    * Returns base64 encoded public key of user current selected account.
--- a/src/@types/casperlabsSigner.d.ts
+++ b/src/@types/casperlabsSigner.d.ts
@@ -17,7 +17,7 @@ interface CasperLabsHelper {
   /**
    * Send Deploy in JSON format message to Signer plugin to sign.
    *
-   * @param messageBase16 the base16 encoded message that plugin received to sign
+   * @param deploy deploy in JSON format
    * @param sourcePublicKey base64 encoded public key used to sign the deploy
    * @param targetPublicKey base64 encoded public key used to sign the deploy
    */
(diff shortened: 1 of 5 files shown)


src/@types/casperlabsSigner.d.ts
```diff
@@ -15,12 +15,13 @@ interface CasperLabsHelper {
   requestConnection: () => void;
   
   /**
-   * send base16 encoded message to plugin to sign
+   * Send Deploy in JSON format message to Signer plugin to sign.
    *
-   * @param messageBase16 the base16 encoded message that plugin received to sign
-   * @param publicKeyBase64 the base64 encoded public key used to sign the deploy, if set, we will check whether it is the same as the active key for signing the message, otherwise, we won't check.
+   * @param deploy deploy in JSON format
+   * @param sourcePublicKey base64 encoded public key used to sign the deploy
```
George-cl: The keys should be in hex format with algorithm prefix. The inline-doc suggests `base64`.

src/@types/casperlabsSigner.d.ts
```diff
@@ -15,12 +15,13 @@ interface CasperLabsHelper {
...
+   * @param targetPublicKey base64 encoded public key used to sign the deploy
    */
-  sign: (messageBase16: string, publicKeyBase64?: string) => Promise<string>;
+  sign: (deploy: JSON, sourcePublicKey: string, targetPublicKey: string) => Promise<JSON>;
```
George-cl: To help cement the idea that we want them to use hex format would it be worth renaming these to:
`sourcePublicKeyHex` and `targetPublicKeyHex`?
Just opinion on this one as the doc will state it explicitly once updated.

Also I can't quite remember but wasn't there an issue with typing `deploy` as `JSON` - I feel like it was set to `JSONTypes`

src/lib/Signer.ts
```diff
@@ -49,19 +49,23 @@ export const getActivePublicKey: () => Promise<string> = () => {
 };
 
 /**
- * send base16 encoded message to plugin to sign
+ * Send Deploy in JSON format message to Signer plugin to sign.
  *
- * @param messageBase16 the base16 encoded message that plugin received to sign
- * @param publicKeyBase64 the base64 encoded public key used to sign the deploy, if set, we will check whether it is the same as the active key for signing the message, otherwise, we won't check.
+ * @param deploy deploy in JSON format
+ * @param sourcePublicKey base64 encoded public key used to sign the deploy
```
George-cl: All references to public keys here should be for the **pub_key_hex** rather than `base64` IMO

src/lib/Signer.ts
```diff
@@ -49,19 +49,23 @@ export const getActivePublicKey: () => Promise<string> = () => {
...
  * @throws Error if haven't connected to CasperLabs Signer browser extension.
  * @throws Error if publicKeyBase64 is not the same as the key that Signer used to sign the message
  */
 export const sign: (
-  messageBase16: string,
-  publicKeyBase64?: string
-) => Promise<string> = (messageBase16: string, publicKeyBase64?: string) => {
-  return window.casperlabsHelper!.sign(messageBase16, publicKeyBase64);
+  deploy: JSON, 
+  sourcePublicKey: string,
```
George-cl, line 64: Same as in the type definition file I suggested it might be worth renaming to specify **hex** in the key var. names

George-cl (changes requested):
Left a few comments - mainly to do with the documentation and naming than anything functional. Happy to discuss any of it further with you.

George-cl (approved):
Cheers - all looks good

PR #84, opened 2021-07-08
Add strong validation before decoding public key

sirormy:


hoffmannjan:
Hi,

thanks for contribution, the problem is your regexp won't cover `SECP256k1` keys which has 66 chars. 
What do you think about `^0(1[0-9a-f]{64}|2[0-9a-f]{66})$` ? 

Cheers!

sirormy:
> Hi,
> 
> thanks for contribution, the problem is your regexp won't cover `SECP256k1` keys which has 66 chars.
> What do you think about `^0(1[0-9a-f]{64}|2[0-9a-f]{66})$` ?
> 
> Cheers!

Yes, I think you are right, code committed, please review, thanks.

Issue #98, opened 2021-08-16
Support public key for transfer target


piotr-dziubecki:
We need to reflect the changes from the core in js sdk.

PR #102, opened 2021-08-27
Changed newTransfer arg from addr hash to hex (next)

hoffmannjan:
Related to: https://github.com/casper-ecosystem/casper-js-sdk/issues/98

### DIFF
--- a/src/lib/DeployUtil.ts
+++ b/src/lib/DeployUtil.ts
@@ -746,7 +746,7 @@ export class ExecutableDeployItem implements ToBytes {
     if (target instanceof CLURef) {
       runtimeArgs.insert('target', target);
     } else if (target instanceof CLPublicKey) {
-      runtimeArgs.insert('target', CLValueBuilder.string(target.toHex()));
+      runtimeArgs.insert('target', target);
     } else {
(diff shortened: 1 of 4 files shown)


src/lib/DeployUtil.ts
```diff
@@ -746,10 +746,7 @@ export class ExecutableDeployItem implements ToBytes {
     if (target instanceof CLURef) {
       runtimeArgs.insert('target', target);
     } else if (target instanceof CLPublicKey) {
-      runtimeArgs.insert(
-        'target',
-        CLValueBuilder.byteArray(target.toAccountHash())
-      );
```
goral09, line 752: It should still support using an account hash. EE will accept that as a target.

hoffmannjan: for the end user it is transparent, they still need to provide `CLUref` or `CLPublicKey` as a target (see line 737).
this is only internal change.

goral09: I see, that's the problem then - Rust library allows for using `AccountHash` type as target. If it's not changing here then that's a separate issue that we don't have to address here.

goral09 (approved):

PR #107, opened 2021-09-23
Casper Provider

chaitanyapotti:
Add support for casper provider

Exposes `CasperServiceByProvider` which extends `CasperServiceByJsonRPC` and takes in a `SafeEventEmitterProvider`.
It uses `ProviderTransport` to communicate with the provider using JsonRPC.

This way JRPC provider can intercept the methods it wants to respond to.
(say: for `account_put_deploy` , jrpc method can take an unsigned deploy, sign it and send to the blockchain)

For now, it intercepts the following methods and performs certain actions:
- `account_put_deploy`
- `info_get_deploy`

PR #109, opened 2021-09-26
Feature/casper provider

hoffmannjan:
Related to #107 

@chaitanyapotti @himanshuchawla009 I changed a little bit structure and creation method of client containing `Provider`, please check.

@chaitanyapotti if you could add an example usage through test eg. deploy creation and then sign through provider (there can be variable containing deploy target URL - we are currently working on running nctl on CI - so in future it can be adjusted).


zie1ony (approved):
Thanks. It's what we wanted :)

PR #113, opened 2021-10-05
Add a bundler config without polyfills

chaitanyapotti:
Currently, casper-js-sdk is our single biggest dependency and it's because it already contains polyfills for node builtins and they're duplicated from our other dependencies. It would be great if there is a build without polyfills built for web.


hoffmannjan (approved):
LGTM

PR #115, opened 2021-10-06
Support historical invalid byte representation of zero

### LINKED ISSUE #103: [Bug] RPC getBlockInfoByHeight "Returned block does not have a matching height"
Hi !

I'm using the getBlockInfoByHeight from the casperServiceByJsonRPC but I'm getting this error on some blocks : 
"Returned block does not have a matching height"

It's not always and not fully reproductible I guess ...

I don't know if there's a cache mechanism with the client used to make the RPC calls ? (I think I've seen that somewhere in a changelog ...)

But I keep getting the error even if I loop over the same height.

Current block impacted on my side : 200003
Block returned : b6900e0b1967de9c527331f77b46970b1a4861a2c1c380c89f1b19c1a7bc8c0b / height : 200766

I'm not using a specific node because I've agregated some nodes behind a reverse proxy.

hoffmannjan:
Added workaround for `U128`, `U256` and `U512` which stores inside the structure bytes used for deserialization, and if structure is unchanged it serializes again to the original bytes. This prevents invalid `bodyHash` problem while validating historical deploys from the network.

fixes #103 @KillianH

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to casper-js-sdk.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 2.6.1
+
+### Fixed
+
+- Added workaround for historical deploys with invalid serialized `0`.
+
 ## 2.6.0
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "casper-js-sdk",
-  "version": "2.6.0",
+  "version": "2.6.1",
   "license": "Apache 2.0",
   "description": "SDK to interact with the Casper blockchain",
   "homepage": "https://github.com/casper-ecosystem/casper-js-sdk#README.md",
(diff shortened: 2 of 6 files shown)


KillianH:
Niceeee I'll try to test it asap

zie1ony (approved):
Looks good. Clever trick.

hoffmannjan:
When @KillianH confirms it is working for him, we can release a new version.

KillianH:
LGTM on mainnet. Parsed all deploys on Mainnet to verify.

Discovered other issues on testnet will report in another issue.

See https://github.com/casper-ecosystem/casper-js-sdk/issues/116

hoffmannjan:
This happens only in new version ?

KillianH:
Yup. Maybe not be related to the same issue. Added a way to test the others broken deploys on testnet on the other issue.

hoffmannjan:
> Yup. Maybe not be related to the same issue. Added a way to test the others broken deploys on testnet on the other issue.

The problem from #116 occurs on every version of JS-SDK that I tested, this is not related. I will look into #116 today.

KillianH:
No problem :) This PR is good. All mainnet deploys & transfers are now working.

PR #121, opened 2021-10-27
Bug : Serialization of RuntimeArgs And folder @types in dist

caspercommunityio:
In my project, I had the following error message :

image

After some researches, I found the error, recompile the sources and test it. It seems to be OK now.

I also updated the webpack config because the folder "@types" was missing in the dist.
It cause me the following error in my project: 
image

Can you rewiew this code ?

Thanks.


hoffmannjan:
Hi @caspercommunityio , LGTM, will give it a second round and should be released this week.
Thanks!

PR #125, opened 2021-11-18
casper-js-sdk 2.7.1

hoffmannjan:
### Fixed

- Added support for mixed case hex representation of public keys introduced in `casper-node 1.4.2`.


goral09 (approved):
We should also update any method that takes a public key (address, hex) and displays it as a text. This should also follow the new scheme that includes the checksum.

hoffmannjan:
> We should also update any method that takes a public key (address, hex) and displays it as a text. This should also follow the new scheme that includes the checksum.

Let's discuss this tomorrow.

PR #135, opened 2022-01-31
casper-js-sdk 2.8.0

hoffmannjan:
- added support to `CLPublicKey` passed as argument to `CLKey` (it gets serialized to `account-hash` when serializing).
- changed `CLKey` method checking stored value type from `isAccount` to `isAccountHash`.
- new `Contracts` class that basically acts as contract client (possible to install WASM, call an entrypoint, query state or dictionary).
- added `getDictionaryItemByName` that uses `state_get_dictionary_item` with `dictionary_name` and `dictionary_item_key`.
- added `sign` and `send` method on `Deploy` class.

### DIFF
--- a/src/lib/Contracts.ts
+++ b/src/lib/Contracts.ts
@@ -115,30 +115,6 @@ export class Contract {
     }
   }
 
-  public async queryContractState(
-    key: string[],
-    stateRootHash?: string,
-    casperClient?: CasperClient
-  ): Promise<CLValue> {
-    const client = casperClient || this.casperClient;
-    if (!client) throw Error(NO_CLIENT_ERR);
-
-    const stateRootHashToUse =
-      stateRootHash || (await client.nodeClient.getStateRootHash());
-
-    const contractData = await this.queryContractData(
-      key,
-      client,
-      stateRootHashToUse
-    );
-
-    if (contractData && contractData.CLValue instanceof CLValue) {
-      return contractData.CLValue.value();
-    } else {
-      throw Error('Invalid stored value');
-    }
-  }
-
   public async queryContractDictionary(
     dictionaryName: string,
     dictionaryItemKey: string,
--- a/src/lib/Contracts.ts
+++ b/src/lib/Contracts.ts
@@ -72,7 +72,9 @@ export class Contract {
   ): Deploy {
     this.checkSetup();
 
-    const contractHashAsByteArray = contractHashToByteArray(this.contractHash!);
+    const contractHashAsByteArray = contractHashToByteArray(
+      this.contractHash!.slice(5)
+    );
 
     const deploy = DeployUtil.makeDeploy(
       new DeployUtil.DeployParams(sender, chainName, 1, ttl),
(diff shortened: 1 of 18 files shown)


src/lib/Contracts.ts
```diff
@@ -19,112 +17,174 @@ export function byteHash(x: Uint8Array): Uint8Array {
...
-    const paymentArgs = RuntimeArgs.fromMap({
-      amount: CLValueBuilder.u512(paymentAmount.toString())
-    });
-
-    const payment = ExecutableDeployItem.newModuleBytes(
-      this.paymentWasm,
-      paymentArgs
+    sender: CLPublicKey,
+    chainName: string,
+    paymentAmount: string,
```
HansWestman74, line 78: Ideally, the developer facing function should not be able to provide a string that cannot be convert to a bigint.

src/lib/Contracts.ts
```diff
@@ -19,112 +17,174 @@ export function byteHash(x: Uint8Array): Uint8Array {
...
+    signingKeys: AsymmetricKey[] = [],
+    ttl: number = DEFAULT_DEPLOY_TTL
+  ): Deploy {
+    this.checkSetup();
+
+    const contractHashAsByteArray = contractHashToByteArray(
+      this.contractHash!.slice(5)
```
HansWestman74, line 85: Add a check that it begins with "hash-"

src/lib/Contracts.ts
```diff
@@ -19,112 +17,174 @@ export function byteHash(x: Uint8Array): Uint8Array {
...
-   * Arguments for Faucet smart contract
-   *
-   * @param accountPublicKeyHash the public key hash that want to be funded
-   */
-  public static args(accountPublicKeyHash: Uint8Array): RuntimeArgs {
-    const accountKey = new CLKey(new CLAccountHash(accountPublicKeyHash));
-    return RuntimeArgs.fromMap({
-      account: accountKey
-    });
+  public async queryContractState(
```
HansWestman74: Should delete this

HansWestman74 (approved):

src/lib/Contracts.ts
hoffmannjan: I just checked and when I provided "ABC" as a payment amount this error got thrown
`Error: invalid BigNumber string (argument="value", value="ABC", code=INVALID_ARGUMENT, version=bignumber/5.1.1)`
from my perspective this is currently enough, there are many places where we want user to provide numeric as a string and adding checks like this at every function would be overkill IMO.

src/lib/Contracts.ts
hoffmannjan: done :)

src/lib/Contracts.ts
hoffmannjan: done, added check in `setContractHash` method :)

George-cl (approved):
LGTM - do you need the `src/types.ts` file if it's empty?

hoffmannjan:
> LGTM - do you need the `src/types.ts` file if it's empty?

nope, removing it, thx :)

PR #136, opened 2022-02-04
casper-js-sdk 2.7.9

hoffmannjan:
### Fixed

- Fixed problems with CLResult when it got created using different version of ts-results.


George-cl (approved):
Needs the unused imports removed from the test but other than great! cheers

PR #141, opened 2022-02-21
Makes targetPublicKey an optional param

George-cl:
Makes the `targetPublicKey` optional and adds doc comment to explain why you would include it.

Also renames `sourcePublicKey` to `signingPublicKey` as it's not just for transfers (where 'source' makes sense) and I think it's just a better name for it.

This is tied into this Signer PR: https://github.com/casper-ecosystem/signer/pull/264

### DIFF
--- a/src/@types/casperlabsSigner.d.ts
+++ b/src/@types/casperlabsSigner.d.ts
@@ -17,16 +17,21 @@ interface CasperLabsHelper {
   requestConnection: () => void;
 
   /**
-   * Send Deploy in JSON format message to Signer plugin to sign.
+   * Send Deploy in JSON format to Signer extension to be signed.
    *
-   * @param deploy deploy in JSON format
-   * @param sourcePublicKeyHex public key in hex format with algorithm prefix. Used to sign the deploy
-   * @param targetPublicKeyHex public key in hex format with algorithm prefix. Used to display hex-formatted address on the UI
+   * @param deploy - deploy in JSON format
+   * @param signingPublicKeyHex - Hex-formatted public key. The corresponding secret key is used to sign the deploy.
+   * @param {string} [targetPublicKeyHex] - Hex-formatted public key.
+   * If the `target` in the deploy is an account hash this can be used to verify it and display the hex-formatted public key in the UI.
+   *
+   * @throws Errors if the Signer extension is not connected.
+   * @throws Errors if signingPublicKey is not available or does not match the Active Key in the Signer.
+   * @throws Errors if targetPublicKeyHex is not the same as the key (or corresponding account hash) that is used as target in deploy.
    */
   sign: (
     deploy: { deploy: JsonTypes },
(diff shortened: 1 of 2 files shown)


src/@types/casperlabsSigner.d.ts
```diff
@@ -17,16 +17,21 @@ interface CasperLabsHelper {
...
+   *
+   * @throws Errors if the Signer extension is not connected.
+   * @throws Errors if signingPublicKey is not available or does not match the Active Key in the Signer.
+   * @throws Errors if targetPublicKeyHex is not the same as the key (or corresponding account hash) that is used as target in deploy.
    */
   sign: (
     deploy: { deploy: JsonTypes },
-    sourcePublicKeyHex: string,
-    targetPublicKeyHex: string
+    signingPublicKeyHex: string,
```
hoffmannjan, line 33: what do you think about making `signingPublicKeyHex` parameter optional too? I believe we shouldn't need it anymore :)

hoffmannjan (approved):

src/@types/casperlabsSigner.d.ts
George-cl: I don't mind so I've raised it with the MAKE team to get their thoughts on it

src/lib/Signer.ts
```diff
@@ -91,29 +91,30 @@ export const getActivePublicKey: () => Promise<string> = () => {
...
- * @throws Error if targetPublicKeyHex is not the same as the key that is used as target in deploy.
+ * @throws Errors if the Signer extension is not connected.
+ * @throws Errors if signingPublicKey is not available or does not match the Active Key in the Signer.
+ * @throws Errors if targetPublicKeyHex is not the same as the key (or corresponding account hash) that is used as target in deploy.
  */
 export const sign: (
   deploy: { deploy: JsonTypes },
-  sourcePublicKey: string,
-  targetPublicKey: string
+  signingPublicKey: string,
```
piotrwitek: I noticed there is a difference in argument names between interface declaration and implementation? 

`targetPublicKeyHex !== targetPublicKey`

Shouldn't consistency be promoted across the board to minimize confusion?

piotrwitek (commented):
question

src/lib/Signer.ts
George-cl: It should match - I missed that cheers

George-cl: Resolved: https://github.com/casper-ecosystem/casper-js-sdk/pull/141/commits/cf48a6f3eb10cd8418fe0bfb24c5af56e9b2e62d

src/@types/casperlabsSigner.d.ts
George-cl: I think we'll leave it as is for now and we can make that change in future, the logic around `signingPublicKeyHex` is slightly more involved.

PR #146, opened 2022-03-17
Adding support to get auction info by block height

kklok:
Adding support to get auction info by block height

{edited out to simplify merge: Also, rm -rf didn't work in windows, so replaced it by a package}

#139


hoffmannjan:
Hi @kklok, the code looks good, thanks for your effort, we will probably get it released this week :)

PR #147, opened 2022-03-18
toJSON on CLValue

hoffmannjan:


src/lib/CLValue/Option.ts
```diff
@@ -144,6 +144,10 @@ export class CLOption<T extends CLValue> extends CLValue {
     return this.data;
   }
 
+  toJSON(): any {
+    return this.isNone() ? 'None' : this.data.unwrap().toJSON();
```
George-cl: Would it be better to return `null` rather than 'None'? Seems closer in meaning to the true sense of None

src/lib/CLValue/Key.ts
```diff
@@ -157,6 +157,10 @@ export class CLKey extends CLValue {
     return this.data;
   }
 
+  toJSON(): any {
```
George-cl: Wouldn't it be worth setting the return type to `string` here?

src/lib/CLValue/Numeric.ts
```diff
@@ -69,6 +69,10 @@ abstract class Numeric extends CLValue {
   value(): BigNumber {
     return this.data;
   }
+
+  toJSON(): any {
```
George-cl: Same here - couldn't it be set to `string`

src/lib/CLValue/Result.ts
```diff
@@ -143,13 +143,21 @@ export class CLResult<T extends CLType, E extends CLType> extends CLValue {
...
    */
   value(): Result<CLValue, CLValue> {
     return this.data;
   }
 
+  /**
+   * Returns JSON representation. If None null will be returned.
+   */
+  toJSON(): any {
+    return this.isError() ? this.data.err : this.data.unwrap().toJSON();
```
George-cl: Do you not need to put a `.toJSON()` on the `this.data.err` because it is a `CLValue` (from https://github.com/casper-ecosystem/casper-js-sdk/blob/4c16adcc5f9125ba8c8d377c091f75abb80e6298/src/lib/CLValue/Result.ts#L136)
Otherwise it won't be formatted properly, unless I'm misunderstanding.

George-cl (changes requested):
Just wee things about specifying return types

src/lib/CLValue/Result.ts
hoffmannjan: in fact in both cases its CLValue, so the check was unnecessary. Thx :+1:

hoffmannjan:
> Just wee things about specifying return types

Thanks @George-cl , fixed the issues.

George-cl (approved):
Cheers Jan, LGTM!

PR #150, opened 2022-03-21
casper-js-sdk 2.9.0

hoffmannjan:
### Added

- Now all stored value contains raw JSON from RPC under `raw` key.
- New method `getValidatorsInfoByBlockHeight` added.

### Changed

- `targetPublicKeyHex` is now optional in `Signer` class.


ihor (approved):
LGTM

PR #151, opened 2022-03-22
Casper-js-sdk 2.9.1

hoffmannjan:
- Got rid of unnecessary instanceof.
- Reverted StoredValue and added optional parameter `{ rawData: boolean }` to `getDictionaryItemByName` and `getDictionaryItemByURef`.


George-cl (approved):
LGTM although I'm not too familiar with the codebase.

ihor (commented):
Looks good to me. I guess we may want to propagate the approach with the `rawData` option to other network calls, but this can be done separately.

hoffmannjan:
> Looks good to me. I guess we may want to propagate the approach with the `rawData` option to other network calls, but this can be done separately.

Indeed, just wanted to provide you working solution ASAP :). It's added to the Roadmap for v2.

Issue #143, opened 2022-03-07
Create roadmap for js-sdk v3 + v2


hoffmannjan:

hoffmannjan:
```
- Phase 1
	- Implement all RPC methods based on the new SDK Specification.
		- All methods should have common configuration argument.
		- It should be possible to get raw JSON from this methods.
		- State Root Hash should always be optional parameter. 
	- Improve the CLValue implementation.
		- Create aliases like `ContractHash` based on Rust implementation.
		- Simplify parts where its possible (eg CLKey).
	- Redesign the whole `DeployUtil`.
		- Implement components that will be needed to build the complex solution.
	
	At this stage it should be possible to call all of the RPC methods as well as manualy construct deploys and send it to the network.

- Phase 2
	- Based on experience with `Web3.js`, `Solana-Web3.js` and `Terra.js` design high-level user-facing API.
		- It should be possible to construct all of the deploys using single methods that provides default values for most popular cases.
		- This API can be discussed with the community.
	- Create complex test suites.
	- Reimplement the `Keys` class so it will be more user friendly (currently it's not easy to find specific methods). 
	- It should be possible to test the SDK over some test-network on CI.

	At this stage we should have well tested MVP and the API proposition that can be discussed.

- Phase 3
	- Implement the user facing API.
	- Release this version as RC and share it with our partners.
	- Based on the tests preapre the v3.0.0

	At this stage we should start annocing the v3 as main release.
	v2 could be marked as deprecated in 3-6 months period.

```

hoffmannjan:
We need a roadmap for v2 and v3 with a milestones. Will work on this in the upcoming sprint.

hoffmannjan:
# casper-js-sdk v3 roadmap sketch

## Main goal
Current [SDK](https://www.npmjs.com/package/casper-js-sdk) is pretty popular library and have a lot of users.
But there is a lot of overenginereed parts that makes end-user confused. We already identified some of them, so v3 should be a good improvement for the end-users (even considering it will contain breaking changes).

We decided to do a rewrite from scratch, as the current version depends on many outdated dependencies and some of the user-facing APIs are not good designed.
So the main goal is simplify the whole library and made it a gold standard for other SDKs.

[...]

PR #152, opened 2022-04-03
signatures verification in validateDeploy()

hoffmannjan:


src/lib/DeployUtil.ts
```diff
@@ -1241,7 +1241,14 @@ export const validateDeploy = (deploy: Deploy): Result<Deploy, string> => {
                   got: ${deploy.hash}.`);
   }
 
-  // TODO: Verify included signatures.
+  deploy.approvals.forEach(({ signer, signature }) => {
+    const pk = CLPublicKey.fromHex(signer);
+    const signatureRaw = decodeBase16(signature.slice(2));
+    const isValid = validateSignature(deploy.hash, signatureRaw, pk);
+    if (!isValid) throw Error('Invalid signature');
```
KillianH: This seems inconsistent with the overall function. It should return an Err() instead of throwing an Error() no ?

src/lib/Keys.ts
```diff
@@ -64,6 +64,24 @@ export function readBase64WithPEM(content: string): Uint8Array {
...
+    return nacl.sign_detached_verify(msg, signature, pk.value());
+  }
+  if (pk.isSecp256K1()) {
+    return secp256k1.ecdsaVerify(
+      signature,
+      sha256(Buffer.from(msg)),
+      pk.value()
+    );
+  }
+  throw Error('Unsupported PublicKey type');
```
KillianH, line 82: Same here ? Not sure if it should throw an error or return false instead ?

test/lib/CasperClient.test.ts
```diff
@@ -131,81 +131,19 @@ describe('CasperClient', () => {
...
-                  parsed: 100000
-                }
-              ]
-            ]
-          }
-        }
-      }
-    };
+    const json = JSON.parse(
+      '{"deploy":{"hash":"1e4111d01748fca01c2986b2f6a576fb105fd8bc93c86e381b5c2b59dd177ef1","header":{"account":"0109791772400ea911e2adcb7569d805da75654fc1360c06f93832f020e13aa0cf","timestamp":"2022-04-03T16:22:46.995Z","ttl":"30m","gas_price":1,"body_hash":"ea0a6bc12489f4ccf0b7564bcacd2918b744b9e4b8cad71d52afd9159f33b108","dependencies":[],"chain_name":"casper-test"},"payment":{"ModuleBytes":{"module_bytes":"","args":[["amount",{"bytes":"0500e40b5402","cl_type":"U512"}]]}},"session":{"Transfer":{"args":[["amount",{"bytes":"0500ba1dd205","cl_type":"U512"}],["target",{"bytes":"01861759c3e71b1953f2be3a92c406a3423fd36ea6a8ff6fd0e71bb39685d68893","cl_type":"PublicKey"}],["id",{"bytes":"01addd020000000000","cl_type":{"Option":"U64"}}]]}},"approvals":[{"signer":"0109791772400ea911e2adcb7569d805da75654fc1360c06f93832f020e13aa0cf","signature":"01ddfc9ae220175a4a81527647b990dbceac97bb684916f5e8b6a40ba416d5174c3717629cd25d914805b7fd62fe0eaa04b943c66f6ad402b25ccca816c0911c03"}]}}'
```
KillianH: Maybe you could squish both tests into a single one with two signature from the two key types ? It will also test the full iteration in the within the for().

src/lib/DeployUtil.ts
hoffmannjan: this is already resolved

src/lib/Keys.ts
hoffmannjan: In practice, this error should never happen, because the `CLPublicKey` is correct when created. 
But in case something really untypical would happen (eg we will change the `CLPublicKey` api) it's better to get an error thrown IMHO.

And it's not returning an `Err` because `Keys` implementation doesn't depend on `ts-results`. This inconsistency is something that we would address in v3 (BTW if you have any suggestions related to v3 please share them [here](https://github.com/casper-ecosystem/casper-js-sdk/issues/143)).

test/lib/CasperClient.test.ts
hoffmannjan: good idea, I changed it that way :+1:. Tests and better code coverage as well as e2e tests for SDK is also something I would address in a near future.

KillianH (approved):

KillianH:
🔥👍

PR #160, opened 2022-06-29
added getBalance

### LINKED ISSUE #155: Update sdk to account for the new query_balance endpoint

hoffmannjan:


src/services/CasperServiceByJsonRPC.ts
```diff
@@ -345,6 +345,23 @@ export class CasperServiceByJsonRPC {
...
+  public async queryBalance(
+    purseIdentifier: string,
+    stateRootHash?: string
+  ): Promise<BigNumber> {
+    return await this.client
+      .request({
+        method: 'query_balance',
+        params: {
+          purse_identifier: {
+            main_purse_under_public_key: purseIdentifier
```
darthsiroftardis: This looks incorrect to me, the purse_identifier arg is being coerced here to the `main_purse_under_public_key` field, which feels like an incorrect implementation of the RPC call. 

Maybe this is me not reading JS very well, and I wouldn't mind going over this in further detail @hoffmannjan

darthsiroftardis (commented):
Need to address, what I believe, is an incorrect implementation

src/services/CasperServiceByJsonRPC.ts
hoffmannjan: You're right! I superficially read the [PR description](https://github.com/casper-network/casper-node/pull/2868) and took it as API spec, my fault!.

I added a new enum `PurseIdentifier` which has variants  matching the `casper-client-rs` implementation. This version of SDK is dump when it comes to type checking an RPC method arguments (thats why I added new argument to `queryBalance` - `purseIdentifierType`) - but that's something that I want to address in `v3` (it will be breaking change for the users).

test/nctl/RPC.test.ts
```diff
@@ -86,6 +87,11 @@ describe('RPC', () => {
     }, 3 * 10000);
   });
 
+  xit('queryBalance', async () => {
```
hoffmannjan, line 90: this tests is disabled, the whole situation needs to be discussed ASAP ([different issue](https://github.com/casper-network/Governance/issues/361) )

George-cl (approved):
Aside from the discussion regarding the disabled test (which I'm not involved in so I don't know  if it's resolved) the actual code changes look good to me.

PR #182, opened 2022-09-04
Extended PublicKey class API with two new methods, allowing to get publicKey tag type or its corresponding signature algo type

piotrwitek:
Problem:
Currently, it's very hard for consumers to get the signature type of a public key stored in a string representation which is the primary way of storage for JS web apps, and also how Signer is representing and providing public keys.

Solution:
Extended PublicKey class API with two new methods, allowing to get the public key tag type or its corresponding signature type. The use case will enable consumers to streamline usage of the many casper-js-sdk APIs requiring signature type as an argument, deriving it directly from a string representation of the public key instead of writing custom and complex logic by consumers that can easily break when `casper-js-sdk` is updated in the future.

Additional info:
- Added test cases for new APIs.
- Fixed incorrect uses of TypeScript types packages convention, causing compile error when executing tests using ts-node, refactored blake.d.ts types to use the supported way following the TypeScript Docs: "A types package is a folder with a file called index.d.ts or a folder with a package.json that has a types field. -- TypeScript Handbook"


piotrwitek:
@hoffmannjan PR discussed on Friday is ready for review. Cheers!

src/lib/CLValue/PublicKey.ts
```diff
@@ -179,4 +179,28 @@ export class CLPublicKey extends CLValue {
 
     return new CLPublicKey(publicKeyHexBytes.subarray(1), publicKeyHexBytes[0]);
   }
+
+  getTag(): CLPublicKeyTag {
+    return this.tag;
+  }
+
+  // TODO: mapping tag to signature algo because of the legacy indentifiers in ./Keys
+  getSignatureAlgorithm(): SignatureAlgorithm {
```
piotrwitek, line 188: This method is to retain compatibility for the legacy identifiers in ./Keys, added special comment for it consistent with other comments for related functionality

hoffmannjan (approved):
LGTM

hoffmannjan:
@piotrwitek thanks. I will merge it this week and we will release new version (most likely by the end of the week). Cheers

piotrwitek:
Awesome, thanks @hoffmannjan 👍 
I saw tests failed for `build 18.x` but the error seems unrelated to my changes. Did you have a look?

hoffmannjan:
Yeah this is something related to one of the dependencies, [will be addressed later](https://github.com/casper-ecosystem/casper-js-sdk/issues/183) :)

piotrwitek:
Thanks @hoffmannjan 👍 When are you planning a new release?

PR #194, opened 2022-10-03
Update CasperClient.ts

omariosman:
I updated a typo in the comment of "loadPublicKeyFromFile" function by changing the word "private key" to "public key".


hoffmannjan:
Good catch @omariosman ! Thanks for this!

hoffmannjan (approved):

PR #198, opened 2022-11-02
this is new request.

ws-jm:
request


hoffmannjan:
We don't merge to release branches as they act as snapshots for already released versions.

Issue #200, opened 2022-11-07
Add support to validate checksums in the mixed-case hex representation of public keys


hoffmannjan:
It's a feature supported by node, but not by the SDK.
SDK should be able to properly validate mixed-case publickeys.

RitaMAllenCA:
Possibly pair programming with Ryo

RitaMAllenCA:
@gyroflaw please assign Jan as a reviewer of this PR.  Thanks

PR #203, opened 2022-11-16
Implement public key checksum validation

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/200


src/lib/ByteConverters.ts
```diff
@@ -145,6 +147,15 @@ export const toBytesVector = <T extends ToBytes>(vec: T[]): Uint8Array => {
...
 
+/**
+ * Use blake2b to compute hash of ByteArray
+ * @param x Byte array of type `Uint8Array` to compute the blake2b hash on
+ * @returns `Uint8Array` buffer of the blake2b hash
+ */
+export function byteHash(x: Uint8Array): Uint8Array {
+  return blake.blake2b(x, null, 32);
+}
+
```
hoffmannjan, line 158: good move with shifting it to this file, you can also switch to use this method `byteHash` in `DeployUtil` file https://github.com/casper-ecosystem/casper-js-sdk/blob/eb5461bbc55fb9adedaa2cf5d819c464a38d38ae/src/lib/DeployUtil.ts#L1355 rather then manually invoking `blake.blake2b`

src/lib/ChecksummedHex.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @returns double sized of Unit8Array
+ */
+const bytesToNibbles = (bytes: Uint8Array): Uint8Array => {
+  let outputNibbles = new Uint8Array();
+  bytes.forEach(byte => {
+    outputNibbles = concat([
+      outputNibbles,
+      Uint8Array.of(byte >>> 4, byte & 0x0f)
+    ]);
+  });
```
hoffmannjan: do you think you can switch from `forEach` to `reduce` with new empty Uint8Array here?

src/lib/ChecksummedHex.ts
```diff
@@ -0,0 +1,98 @@
...
+  return outputNibbles;
+};
+
+const bytesToBitsCycle = (bytes: Uint8Array) => {
+  const output: boolean[] = [];
+  for (let i = 0, k = 0; i < bytes.length; i++)
+    for (let j = 0; j < 8; j++)
+      output[k++] = ((bytes[i] >>> j) & 0x01) === 0x01;
+
+  return output;
```
hoffmannjan, line 32: nice and optimal way :+1:

src/lib/ChecksummedHex.ts
```diff
@@ -0,0 +1,98 @@
...
+ * - Uses Blake2b hashes rather than Keccak
+ * - Uses hash bits rather than nibbles
+ * @param input Uint8Array to generate checksummed hex string
+ * @returns checksummed hex presentation string of input
+ */
+export const encode = (input: Uint8Array): string => {
+  const inputNibbles = bytesToNibbles(input);
+  const hashBits = bytesToBitsCycle(byteHash(input)).values();
+  let hexOutputString = '';
+  inputNibbles.forEach(nibble => {
```
hoffmannjan: again, maybe we can use one `const` and `reduce` here rather than `forEach` and `let`?

src/lib/ChecksummedHex.ts
```diff
@@ -0,0 +1,98 @@
...
+ * similar to scheme in [EIP-55](https://eips.ethereum.org/EIPS/eip-55).
+ * Key differences:
+ * - Works on any length of (decoded) data up to `SMALL_BYTES_COUNT`, not just 20-byte addresses
+ * - Uses Blake2b hashes rather than Keccak
+ * - Uses hash bits rather than nibbles
+ * For backward compatibility: if the hex string is all uppercase or all lowercase, the check is
+ * skipped.
+ * @param input string to check if it is checksummed
+ * @returns true if input is checksummed
+ */
```
hoffmannjan, line 81: very nice documentation @gyroflaw 🥇

test/lib/ChecksummedHex.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      '020310Ae2b7191bA40b97e0531CE84554345782beE5Fd4574870359AfB9EE7eB871E',
+      '0203d7F45A5a34dD5b0131afae10aF69A9D1aF16FA2BcC35fD76d0402A1d2fFF461f',
+      '0202848e3889e6F6806d2e163BbEaFC2818C9F1d8984660A2e135e2BcD72fA9c2c4B'
+    ];
+
+    validInputs.map(isChecksummed).should.not.include(false);
+    const invalidInputs = validInputs.map(i => i.toLowerCase());
+    invalidInputs.map(isChecksummed).should.not.include(true);
+  });
+});
```
hoffmannjan, line 82: awesome ! ❤️

src/lib/CLValue/PublicKey.ts
```diff
@@ -165,7 +176,7 @@ export class CLPublicKey extends CLValue {
 
   /**
    * Tries to decode PublicKey from its hex-representation.
-   * The hex format should be as produced by PublicKey.toAccountHex
+   * The hex format should be as produced by CLPublicKey.toHex
    * @param publicKeyHex
    */
   static fromHex(publicKeyHex: string): CLPublicKey {
```
hoffmannjan: maybe an optional second parameter here to force checksumed variant of pubkey? then it can throw an error that the value might be invalid because we can't validate it? and for dapp develop it will be handy to just `try catch` and handle in his app.

src/lib/CLValue/PublicKey.ts
```diff
@@ -120,8 +125,14 @@ export class CLPublicKey extends CLValue {
     return this.tag === CLPublicKeyTag.SECP256K1;
   }
 
-  toHex(): string {
-    return `0${this.tag}${encodeBase16(this.data)}`;
+  toHex(checksummed = false): string {
```
hoffmannjan: hm, I think it can be `true` by default, because if `fromHex` accepts also non-checksummed variant it shouldn't break anything for anybody. but we will add a note that next release will contain this change.

hoffmannjan (changes requested):
Amazing job @gyroflaw! Congratulations 🏅! I only found a few small issues that we might address :). And also please add to `CHANGELOG.md` the info in a format

```
## [Next-Release]

### Added

-

### Fixed
-

### Changed
-
```

hoffmannjan (approved):
Amazing work Ryo! I added a CHANGELOG proposition, please look at it and if agree you can merge this PR.

PR #174, opened 2022-08-02
Added mote/cspr converters

dylanireland:
Added Mote to CSPR and CSPR to mote converters in Conversions.ts

PR #206, opened 2022-11-24
Update conversions export

gRoussac:
It seems this PR #174 added functions but did not export them module wise. 

I have a test failing, not sure if related 
```
  1 failing

  1) CasperClient
       should create a HK wallet and derive child account correctly:
     Error: error:0308010C:digital envelope routines::unsupported

```


hoffmannjan (approved):
Good catch! thanks for this

Issue #167, opened 2022-07-23
[BUG] Cannot read properties of undefined / Invalid Deploy


KillianH:
**Describe the bug**
Following #116 Multiple deploys are still unparsable with the SDK

**To Reproduce**

```js
const { CasperClient } = require('casper-js-sdk');

const casperClient = new CasperClient('http://rpc.testnet.casperholders.com/rpc');

async function test() {
  const deployArrayBug = [
    '1eea0ed5ab843af2c5cb189bb13b3094d07d9389e2e04417e26c8c2a53004a6b',
    'e2bed34a3b8ea4af52c39f0baef6492ae05cd59da3883328e34eb8b943b15425',
    '3b333c2fadf34ef160ad36e96d34d0223b3ecc1f12fd6370bc1728e9129f0e05',
    '588ef35f526747e6f66a2ae49bad8ac0022409ed78746b7b9610777751453b83',
    '886c87b4373ee5bfdcc4047c12871de728cbae8f2e5e87ca55e1d6d1ad3d943c',
    '9bcf8c15f111be6adb7c5c7601e47de6ec0ecfa226e471ee8d5728cc9e1fb1dd',
    '0fc55ccbdb0c9a145e0517c08c4068b43a942e0147f6393d89a4690cb788087c',
    '691ee9e6787acae4eda2ed0b3731615a2a5b33f19e68ba25c0923a5d45a986b6',
    '7bd9a73809b4b8227306e4ac216edd38c590c81628ea04147fc757d0a8927d00',
    'af039ea1154e9ba49b723da9226ae65ec4e8c78370e4ef2786654aa10384c0eb',
    'b5e5c3138f67212e850d4d003f2ef8e46271c9f778a82c0b25a942350646026c',
    '5ae48d77a1deff3f857c284658710facc95553ab7af9e8e48e64a96f51a63c67',
    '3e3718099023223021501d6f2f8f201f2815052a7dfc699814e93ed5f0b4ffa8',
    '398c13d4226e883f30ca54cea6c6cefc88528e3dcc18013b79894ff3e8790b01',
    '1e32504f887df2e33db07cd9b48567290614c943f3518c06b7787d469b0b726b',
    '4a390833cc0c8c02c953332135f9521da4d2141b105297c78aaabefe461e51d3',
    '5887ca0e0976babf3534f2f7d4c68281f60a37f4736c9d82f8f50a16361a687a',
[...]

PR #209, opened 2022-12-05
Fix failed deploy parse

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/167

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "~build": "tsc --watch",
     "test:rpc": "TS_NODE_FILES=true mocha -r ts-node/register test/rpc/*.test.ts --timeout 5000000",
     "test": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/*.test.ts src/lib/CLValue/*.test.ts",
-    "test2": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/CasperClient.test.ts",
     "test:coverage": "nyc npm run test",
     "docs": "typedoc --out docs/ src",
     "publish-doc": "yarn run docs && gh-pages -d docs -r https://github.com/casper-ecosystem/casper-client-sdk-docs"
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "~build": "tsc --watch",
     "test:rpc": "TS_NODE_FILES=true mocha -r ts-node/register test/rpc/*.test.ts --timeout 5000000",
     "test": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/*.test.ts src/lib/CLValue/*.test.ts",
+    "test2": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/CasperClient.test.ts",
     "test:coverage": "nyc npm run test",
     "docs": "typedoc --out docs/ src",
     "publish-doc": "yarn run docs && gh-pages -d docs -r https://github.com/casper-ecosystem/casper-client-sdk-docs"
--- a/src/lib/CLValue/AccountHash.ts
+++ b/src/lib/CLValue/AccountHash.ts
@@ -37,12 +37,14 @@ export class CLAccountHashBytesParser extends CLValueBytesParsers {
     bytes: Uint8Array
   ): ResultAndRemainder<CLAccountHash, CLErrorCodes> {
     if (bytes.length < ACCOUNT_HASH_LENGTH) {
-      return resultHelper(Err(CLErrorCodes.EarlyEndOfStream));
+      return resultHelper<CLAccountHash, CLErrorCodes>(
+        Err(CLErrorCodes.EarlyEndOfStream)
+      );
     }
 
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
     "~build": "tsc --watch",
     "test:rpc": "TS_NODE_FILES=true mocha -r ts-node/register test/rpc/*.test.ts --timeout 5000000",
     "test": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/*.test.ts src/lib/CLValue/*.test.ts",
+    "test2": "TS_NODE_FILES=true mocha -r ts-node/register test/lib/CasperClient.test.ts",
```
hoffmannjan: we can get rid of this :)

src/lib/CLValue/AccountHash.ts
```diff
@@ -37,12 +37,14 @@ export class CLAccountHashBytesParser extends CLValueBytesParsers {
...
-      return resultHelper(Err(CLErrorCodes.EarlyEndOfStream));
+      return resultHelper<CLAccountHash, CLErrorCodes>(
+        Err(CLErrorCodes.EarlyEndOfStream)
+      );
     }
 
     const accountHashBytes = bytes.subarray(0, ACCOUNT_HASH_LENGTH);
     const accountHash = new CLAccountHash(accountHashBytes);
-    return resultHelper(Ok(accountHash));
+    return resultHelper(Ok(accountHash), bytes.subarray(ACCOUNT_HASH_LENGTH));
```
hoffmannjan, line 47: good catch!

src/lib/CLValue/Map.ts
```diff
@@ -180,6 +191,10 @@ export class CLMap<K extends CLValue, V extends CLValue> extends CLValue {
     return this.data;
   }
 
+  has(k: K): boolean {
+    return this.get(k) !== undefined;
+  }
+
```
hoffmannjan, line 197: nice feature :)

src/lib/RuntimeArgs.ts
```diff
@@ -34,20 +34,27 @@ export class NamedArg implements ToBytes {
...
     Array.from(_arr, ([key, value]) => {
       const val = CLValueParsers.fromJSON(value);
       return [key, val.unwrap()];
     })
   );
+
+  if (parsed.size !== Array.from(_arr).length)
+    throw Error(`Duplicate key exists.`);
+
+  return parsed;
```
hoffmannjan, line 57: in fact, this is acceptable to have two args with the same name - from node perspective they are represented as a vector of pairs, so for example if you have `[("a", 1), ("b", 2)]` `get_named_arg("a")` will return `1`. In our case we're stroring it as `Map` which prevents us from having duplicates. For now I think it's ok how it works, but we need to take info consideration that in v3 we will need to switch from native `Map` to our own implementation of list of pairs :).

hoffmannjan (approved):
nice work Ryo! did you contracted KillianH to check if it works for him?

src/lib/RuntimeArgs.ts
gyroflaw: For about this, the [deploy on testnet](https://testnet.cspr.live/deploy/244d731044e1b7059507b9191a041844e93c8afa8789f03b5c5632ea85a7e286)
has 2 args for `store_tuple2` entry point.
But as you can see below
image
the values named `value` have two values: `valueforentrypointstore_tuple2` for string type and `123` for U512.
So we need to check if arguments shouldn't have duplicated keys.

hoffmannjan:
@RitaMAllenCA I will merge this after we will prepare a new version to release for tomorrow. Don't wan't to include it in this week release as it contains pretty big changes and that not the best moment to make it public.

RitaMAllenCA:
@hoffmannjan Sure - makes sense.  You are releasing V2.10.1 today?
Will we wait for 3.0 to release the feature since it is breaking?

KillianH:
> @hoffmannjan Sure - makes sense. You are releasing V2.10.1 today? Will we wait for 3.0 to release the feature since it is breaking?

I'm not in a hurry for the fix since I now use GoLang to parse every deploys on the networks. Still think that the SDK should parse any deploys sent to the network even if some are failed with strange values. Good job everyone :)

src/lib/RuntimeArgs.ts
hoffmannjan: But this is not how the node acts. as I said node will accepts two arguments with the same name, it will just use the first one provided. in SDK we should replicate this behaviour, just wanted you to be aware about this, this will require use of list of pairs, instead of `Map`. But not for the 2.x.x, we will do this in 3.x

hoffmannjan:
> @hoffmannjan Sure - makes sense. You are releasing V2.10.1 today? Will we wait for 3.0 to release the feature since it is breaking?

no, 3.0 it's totally different thing, any of the PR's is not related to v3. works on v3 will probably begin next year.

RitaMAllenCA:
thank you Jan

Issue #175, opened 2022-08-02
Extended test scenarios for JS-SDK


cspramit:
Add extended, stateful test scenarios that deploys some WASM over the RPC, query dictionaries, do some transfers between accounts etc.

RitaMAllenCA:
@gyroflaw please stop work on this and move to the ticket in the 11/15 sprint.

PR #211, opened 2022-12-07
Extend test case for `CasperServiceByJsonRPC`

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/175

TODO:
- [x] Native token transfer
- [x] ERC20 token deploy and query contract data


src/lib/Contracts.ts
```diff
@@ -138,7 +137,7 @@ export class Contract {
     path: string[] = [],
     casperClient?: CasperClient,
     stateRootHash?: string
-  ): Promise<StoredValue> {
+  ): Promise<any> {
```
hoffmannjan, line 140: this is pain, but for now we won't do anything better, right?

test/rpc/RPC.test.ts
```diff
@@ -80,12 +96,9 @@ describe('RPC', () => {
     assert.equal(stateRootHash.length, 64);
   });
 
-  it('state_get_balance', async () => {
+  xit('state_get_balance', async () => {
```
hoffmannjan: why its commented out?

test/rpc/RPC.test.ts
```diff
@@ -139,14 +155,123 @@ describe('RPC', () => {
...
+      faucetKey.publicKey
+    );
     const [prefix, value, suffix] = uref.split('-');
     expect(uref).to.be.equal(faucetMainPurseUref);
     expect(prefix).to.be.equal('uref');
     expect(value.length).to.be.equal(64);
     expect(suffix.length).to.be.equal(3);
   });
 
+  xit('should transfer native token by session', async () => {
```
hoffmannjan: and this too?

test/rpc/RPC.test.ts
```diff
@@ -164,10 +289,8 @@ describe('RPC', () => {
   //   const eraSummary = await client.getEraInfoBySwitchBlockHeight(10);
```
hoffmannjan: can we uncomment this code and make it working? I mean to check if we can find the deploys that was done before  in the erc20 section?

hoffmannjan (changes requested):
Overall looking good, I would just expend a few test after erc20 token was deployed so we can check few other SDKs methods ☺️

src/lib/Contracts.ts
gyroflaw: yeah, you are right

hoffmannjan (approved):
LGTM :+1:

PR #219, opened 2022-12-21
Export Result in sdk

gRoussac:
Hi, those lines in DeployUtil

```
export declare const deployFromJson: (json: any) => Result<Deploy, Error>;

export declare const validateDeploy: (deploy: Deploy) => Result<Deploy, string>;
```

return `Result` as type

and this type is not exported so in my app if I need to do

`const signedDeploy: Result<DeployUtil.Deploy, Error> = DeployUtil.deployFromJson(signedDeployToJson);`

then I need to import in my app

`"ts-results": "npm:@casperlabs/ts-results@^3.3.4",`

which is not ideal.

It's a bit sensitive, please check that it does not break anything.

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -1,9 +1,9 @@
+import { Result } from 'ts-results';
 import * as Contracts from './Contracts';
 import * as DeployUtil from './DeployUtil';
 import * as Keys from './Keys';
 import * as Serialization from './Serialization';
 import * as Signer from './Signer';
-import { Result } from 'ts-results';
 export * from './CLValue';
 export * from './StoredValue';
 export * from './RuntimeArgs';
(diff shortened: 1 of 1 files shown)


src/lib/index.ts
```diff
@@ -3,11 +3,12 @@ import * as DeployUtil from './DeployUtil';
 import * as Keys from './Keys';
 import * as Serialization from './Serialization';
 import * as Signer from './Signer';
+import { Result } from 'ts-results';
```
gyroflaw, line 6: NIT: How about moving this line to the top of the file since it is an external library import :)

gyroflaw (approved):

src/lib/index.ts
gRoussac: moved

gyroflaw: LGTM 👍

hoffmannjan (approved):

PR #236, opened 2023-02-07
AccountHash should be possible to serialize as ByteArray

### LINKED ISSUE #224: AccountHash support
**Describe the bug**
AccountHash CLType isn't supported

https://github.com/casper-ecosystem/casper-js-sdk/blob/7da7c50/src/lib/CLValue/AccountHash.ts#L66

The Account Hash CLType is missing the toBytes() method. This is needed in order to support CasperLabs private auction smart contract.

Can you provide a way to insert an account hash with the sdk into a deploy ?

hoffmannjan:
Needed to create an alias for `AccountHash` so it can be serialized with same CLTYPE byte as `ByteArray`

~~Didn't wanted to change a lot of logic so I added workaround to deal with aliases - in v3 we will need to rethink this concept :).~~

Scope of this PR is a little bit bigger than initial:
- changed namings in some internals of `CLValue` and `CLType`
- added easier method of creating aliases
- changed `ifs` in `matchByteParserByCLType` to switch statement


hoffmannjan:
I was sick because it was temp solution - so I spend a little extra time and made it final solution - also I'm proud that we're adding new features with more removed lines than added <3

src/lib/CLValue/AccountHash.test.ts
```diff
@@ -8,4 +9,18 @@ describe('CLAccountHash', () => {
...
+    const hash = new CLAccountHash(expectedBytes);
+    const bytes = CLValueParsers.toBytes(hash).unwrap();
+
+    const fromBytes = CLValueParsers.fromBytes(
+      bytes,
+      new CLAccountHashType()
+    ).unwrap();
+
+    expect(bytes).to.deep.eq(expectedBytes);
+    expect(fromBytes).to.deep.eq(hash);
```
gRoussac:
```suggestion
  it('toBytes() / fromBytes() do proper bytes serialization', () => {
    const expectedBytes = Uint8Array.from(Array(32).fill(42));
    const expectedHash = new CLAccountHash(expectedBytes);
    
    const bytes = CLValueParsers.toBytes(hash).unwrap();
    const hash = CLValueParsers.fromBytes(
      bytes,
      new CLAccountHashType()
    ).unwrap();

    expect(bytes).to.deep.eq(expectedBytes);
    expect(hash).to.deep.eq(expectedHash);
```

src/lib/CLValue/AccountHash.ts
```diff
@@ -12,19 +9,16 @@ import {
...
-    return ACCOUNT_HASH_ID;
-  }
+// AccountHash is an alias, not a fully functional CLType so uses the same CLTypeTag as ByteArray
+export class CLAccountHashType extends CLByteArrayType {
+  linksTo = ACCOUNT_HASH_ID;
 
-  toJSON(): string {
-    return this.toString();
+  constructor() {
+    super(32);
```
gRoussac:
```suggestion
    const size = 32;
    super(size);
```

gRoussac: This 32 looks like CL_BYTE_ARRAY_MAX_LENGTH, if this can be whether imported here maybe, or have it in one place defined maybe.

src/lib/CLValue/Bool.ts
```diff
@@ -12,16 +12,8 @@ import {
 import { BOOL_ID, CLTypeTag } from './constants';
 
 export class CLBoolType extends CLType {
-  linksTo = CLBool;
+  linksTo = BOOL_ID;
```
gRoussac: Using _ID suffix is a bit confusing, it seems to indicate some kind of numerical identifier while actually it's a type. I know naming is hard but this does not seem an ID but a Type.

src/lib/CLValue/constants.ts
```diff
@@ -71,6 +71,8 @@ export enum CLTypeTag {
   PublicKey = 22
 }
 
+export const ALIAS_ACCOUNT_HASH_TO_BYTE_ARRAY = -1;
```
gRoussac: Can you please comment this one ? what does this mean exactly to receive -1 ?

Where do you use it please ? Either github tricks me, either it is not used.

src/lib/CLValue/utils.ts
```diff
@@ -156,68 +158,71 @@ export const matchTypeToCLType = (type: any): CLType => {
 export const matchByteParserByCLType = (
   val: CLType
 ): Result<CLValueBytesParsers, string> => {
-  if (val.tag === CLTypeTag.Bool) {
+  if (val.linksTo === ACCOUNT_HASH_ID) {
```
gRoussac: This file looks like a big switch whether than successive if statements. I don't believe this style would pass a code smell sniffer like Sonarcloud. I would suggest to ask for a subscription at Sonarcloud to Ed and pass the whole plugin into their analyser. That much retunrs in a function would probably not be accepted.

src/lib/CLValue/Bool.ts
gRoussac: I would suggest to grep and replace _ID by _TYPE but this could be a matter of taste.

src/lib/CLValue/Abstract.ts
```diff
@@ -34,14 +34,17 @@ export const resultHelper = <T, E>(
...
+  abstract linksTo: string;
   abstract tag: CLTypeTag;
-
   toBytes(): Uint8Array {
     return Uint8Array.from([this.tag]);
   }
+  toString(): string {
+    return this.linksTo;
+  }
+  toJSON(): unknown {
```
gRoussac, line 45: Just curious how could this return an `unknown` type if this returns toString in any case.

src/lib/CLValue/AccountHash.ts
hoffmannjan: replaced for `ACCOUNT_HASH_LENGTH` that was defined in `constants`

src/lib/CLValue/Abstract.ts
hoffmannjan: its beeing overwritten in some places like for example in [`CLMap`](https://github.com/casper-ecosystem/casper-js-sdk/blob/dev/src/lib/CLValue/Map.ts#L63-L70)

src/lib/CLValue/utils.ts
hoffmannjan: I changed it to `switch`, there is still one place where I needed to use `ifs` but hopefully this can be refactored someday...

src/lib/CLValue/Bool.ts
hoffmannjan: done :) thanks!

src/lib/CLValue/Abstract.ts
```diff
@@ -34,14 +34,17 @@ export const resultHelper = <T, E>(
 };
 
 export abstract class CLType {
-  abstract toString(): string;
-  abstract toJSON(): any;
-  abstract linksTo: any;
+  abstract linksTo: string;
```
gyroflaw, line 37: I wanted to aske before, what's the purpose of linksTo?

src/lib/CLValue/Abstract.ts
gRoussac: This seems to me unclear how CLMap::toJSON() can return any type ? isn't it a string too ?

I would remove that `unknow` return type here anyway and let it infer the type, it's a bit confusing tbh. `unknow` is mainly forbidden to use or when you don't have the choice when casting as, but I would not use it in a function definition.

src/lib/CLValue/AccountHash.ts
```diff
@@ -10,21 +7,18 @@ import {
...
 
-export class CLAccountHashType extends CLType {
-  linksTo = CLAccountHash;
-  tag = -1;
-
-  toString(): string {
-    return ACCOUNT_HASH_ID;
-  }
+// AccountHash is an alias, not a fully functional CLType so uses the same CLTypeTag as ByteArray
+export class CLAccountHashType extends CLByteArrayType {
```
gyroflaw, line 17: I think this is the best way without breaking any 👍

gRoussac (approved):

src/lib/CLValue/Bool.ts
```diff
@@ -61,19 +53,4 @@ export class CLBool extends CLValue {
...
-      return resultHelper(Err(CLErrorCodes.EarlyEndOfStream));
-    }
-    if (bytes[0] === 1) {
-      return resultHelper(Ok(new CLBool(true)), bytes.subarray(1));
-    } else if (bytes[0] === 0) {
-      return resultHelper(Ok(new CLBool(false)), bytes.subarray(1));
-    } else {
-      return resultHelper(Err(CLErrorCodes.Formatting));
-    }
-  }
```
gyroflaw, line 78: 👍

src/lib/CLValue/Abstract.ts
hoffmannjan: no, `CLMap::toJSON()` will return object, I send you an example in a link above.
In most cases it returns string, but this is not a rule. Sometimes its overwritten to return proper `cl_type`.

For me its more adequate to use `unknown` here instead of any, look at this documentation https://devblogs.microsoft.com/typescript/announcing-typescript-3-0/#the-unknown-type

src/lib/CLValue/Abstract.ts
hoffmannjan: it was unused, but when designing we plan to use it for something else. so mainly a legacy thing, and now I used it to distinguish `tag` which is used to prepend byte that describes type from something that is mainly used internally to check the type

src/lib/CLValue/constants.ts
hoffmannjan: it was a temp thing, if you look at the previous implementation `AccountHash` was only type used internally which was not possible to serialize. Now it's not used anymore, I removed it, thanks!

src/lib/CLValue/Abstract.ts
gyroflaw: IMO, the return type of `toJSON` method should be `any` or `unknow` because there is `CLType::Any`.
But I think the `toJSON` method returns this object in general.

```json
{
  "bytes": <hex string representation of bytes>,
  "parsed": <parsed data:unknown>,
  "cl_type": <string represent of CLType>
}
```
So I think the return type of the toJSON should be

```typescript
type ToJSON = {
 bytes: string;
 data: unknown;
 cl_type: CLType Enum string
}
```

gyroflaw (commented):
All changes LGTM 👍 , and just a minor address regarding return type of `toJSON` method of `CLType` class. I think we should consider this change since it is basic class of the SDK.

src/lib/CLValue/Abstract.ts
hoffmannjan: I got your point, but this is `toJSON` on `CLType` so it only returns `cl_type` field :).  And [here it is combined](https://github.com/casper-ecosystem/casper-js-sdk/blob/dev/src/lib/CLValue/Abstract.ts#L79-L84) into your `ToJSON` type

gyroflaw: To many `toJSON` methods, just confused.

gyroflaw (approved):
I think the remain thing is just updating PR description since a lot of changes are made here, :)

PR #238, opened 2023-02-10
Fix ContractPackageJson deserialize error

gyroflaw:
Fixes https://github.com/CasperLabs/Metacask_Trakr/issues/5
- Also fixed eslint-prettier rule (removed eslint-plugin-pretter and used eslint-config-pretter), 
- replaced lodash module, 
- removed unused module (@types/express)


package.json
```diff
@@ -46,9 +46,8 @@
   },
   "devDependencies": {
     "@types/chai": "^4.1.7",
-    "@types/express": "^4.17.0",
     "@types/humanize-duration": "^3.18.1",
-    "@types/lodash.omit": "^4.5.7",
+    "@types/lodash": "^4.14.191",
```
hoffmannjan, line 51: do we need the whole lodash? or is it anyway stripped when bundling?

hoffmannjan (approved):
Overall it LGTM, only one small question related to `lodash` :)

package.json
gyroflaw: TBH, we don't need whole lodash but since it used by importing sub modules like [here](https://github.com/casper-ecosystem/casper-js-sdk/blob/ca0c22a12472d7bf14a9e919fce989cc4ff4de12/test/lib/StoredValue.test.ts#L1-L2)

```ts
import mapKeys from 'lodash/mapKeys';
import camelCase from 'lodash/camelCase';
```

don't need to worry about size. I tried to add [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin) but this requires lots of modification on webpack config. So please keep in mind that every lodash packages must be imported as sub-module.<br />
I will add [eslint-plugin-lodash](https://www.npmjs.com/package/eslint-plugin-lodash) to prevent import whole lodash.

src/lib/StoredValue.ts
```diff
@@ -261,8 +261,8 @@ export class GroupsJson {
   @jsonMember({ name: 'group', constructor: String })
   public group: string;
```
hoffmannjan, line 262: you missed this part!

hoffmannjan (changes requested):

Issue #239, opened 2023-02-10
Consider replacing `Secp256k1` keypair generation module.


gyroflaw:
Found that we're using [eccrypto](https://www.npmjs.com/package/eccrypto) module to generate `secp256k1` keypair even we can generate `secp256k1` keypair using `ethereum-cryptography/secp256k1` module.

Suggested change:

from
```ts
  public static new() {
    const privateKey = eccrypto.generatePrivate();
    const publicKey = Uint8Array.from(eccrypto.getPublicCompressed(privateKey));
    return new Secp256K1(publicKey, privateKey);
  }
```

to 
```ts
  public static new() {
    const privateKey = secp256k1.createPrivateKeySync();
    const publicKey = secp256k1.publicKeyCreate(privateKey);
    return new Secp256K1(publicKey, privateKey);
  }
```

_Originally posted by @gyroflaw in https://github.com/casper-ecosystem/casper-js-sdk/issues/180#issuecomment-1425771510_

hoffmannjan:
Yeah, just please take into consideration that it needs to work both in browser and nodejs env :)

PR #252, opened 2023-02-17
Use noble scope libraries for key generation and hash

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/239


hoffmannjan:
@RitaMAllenCA I'm checking this PR but want to run extensive compatibility test so we won't regret releasing this 😅

hoffmannjan:
@gyroflaw could you take a look and fix the conflicts? :)

PR #278, opened 2023-03-30
Fix nightly test error

gyroflaw:
Fix nightly test [error](https://github.com/casper-ecosystem/casper-js-sdk/actions/runs/4548894282)


hoffmannjan (approved):
LGTM

Issue #199, opened 2022-11-03
[Feature Request] Support configurable timeout for requests


cy-cdc:
Hi, I notice that the implementation of the Transport class ultimately supports passing in a [request timeout value as a parameter](https://github.com/casper-ecosystem/casper-js-sdk/blob/673b176276960d740355f62462b736b71fc4306f/src/services/ProviderTransport.ts#L52). 
However, this does not seem to be exposed any where in the options to allow us to configure this value. Would be nice to have this, cheers!

gyroflaw:
@cy-cdc , could you please share the code snippet you're using?

PR #284, opened 2023-04-03
Add timeout params

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/199

TODO:

- [ ] Add test case for `ProviderTransport` class


hoffmannjan (approved):

src/services/CasperServiceByJsonRPC.ts
```diff
@@ -303,36 +303,46 @@ export class CasperServiceByJsonRPC {
   /**
    * Get information about a deploy using its hexadecimal hash
    * @param deployHashBase16 The base-16 hash of the deploy
+   * @param timeout timeout for the request
    * @returns A `Promise` that resolves to a `GetDeployResult`
    */
   public async getDeployInfo(
-    deployHashBase16: string
+    deployHashBase16: string,
+    timeout?: number
```
hoffmannjan: can we have here an object like 
`{ timeout?: value }`? That way it will be easier to expend in future.

hoffmannjan (changes requested):
Just one small change that will make it future proof

src/services/CasperServiceByJsonRPC.ts
```diff
@@ -303,36 +315,46 @@ export class CasperServiceByJsonRPC {
   /**
    * Get information about a deploy using its hexadecimal hash
    * @param deployHashBase16 The base-16 hash of the deploy
+   * @param timeout timeout for the request
```
hoffmannjan: you need to update the docs to include `props` :)

hoffmannjan (changes requested):
last change in docs and we're good 👍

hoffmannjan (approved):

Issue #292, opened 2023-04-28
GitHub CI script doesn't return the latest casper-node version


gyroflaw:
The CI script should return [v1.4.15](https://github.com/casper-network/casper-node/releases/tag/v1.4.15) but returned v1.4.9
Check here: https://github.com/casper-ecosystem/casper-js-sdk/actions/runs/4829494304/jobs/8604602661

gyroflaw:
Current command to fetch latest

```bash
curl -s https://api.github.com/repos/casper-network/casper-node/tags | jq -r '.[].name' | grep 'v*' | sort -r | head -n1
```

Suggest

```bash
curl -s https://api.github.com/repos/casper-network/casper-node/tags | jq -r '.[].name' | grep 'v*' | sort -V | sed 's/_$//' | tail -n1
```


Please remove `head -n1` or `tail -n1` to sorted tags list

PR #297, opened 2023-05-19
Fix fetch the latest version of node command

gyroflaw:
Fix https://github.com/casper-ecosystem/casper-js-sdk/issues/292

### DIFF
--- a/.github/workflows/ci-casper-client-sdk.yml
+++ b/.github/workflows/ci-casper-client-sdk.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Get Latest Casper-Node Tag Version
         run: |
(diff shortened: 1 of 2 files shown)


gyroflaw:
Current command to fetch latest

```bash
curl -s https://api.github.com/repos/casper-network/casper-node/tags | jq -r '.[].name' | grep 'v*' | sort -r | head -n1
```

Suggest

```bash
curl -s https://api.github.com/repos/casper-network/casper-node/tags | jq -r '.[].name' | grep 'v*' | sort -V | sed 's/_$//' | tail -n1
```


Please remove `head -n1` or `tail -n1` to sorted tags list

hoffmannjan (approved):
👍

Issue #313, opened 2023-06-14
CI Actions are failed


gyroflaw:
There are some issues in node setup ci script and needs to check

PR #316, opened 2023-06-16
Fix CI issue

gyroflaw:
Fixes https://github.com/casper-ecosystem/casper-js-sdk/issues/313


gyroflaw:
There was an issue with S3

PR #339, opened 2023-07-18
Fix get state root hash RPC

KillianH:
State root hash RPC signature have changed with 1.5 https://docs.casper.network/developers/json-rpc/json-rpc-informational/#chain-get-state-root-hash

So the current implementation doesn't work with testnet 1.5.


gyroflaw (approved):
LGTM 👍, thank you

Issue #343, opened 2023-07-25
Update `CasperServiceByJsonRPC` class


gyroflaw:
### Context

There were some changes in JSON RPC request from casper-node 1.5 version

PR #344, opened 2023-07-25
Update RPC methods

gyroflaw:
Fixes https://github.com/casper-ecosystem/casper-js-sdk/issues/343

Updated JSON RPC requests based on the [schema](http://135.181.140.222:8888/rpc-schema), and [OpenRPC playground](https://playground.open-rpc.org/)


hoffmannjan (commented):
Looking good, but I would add tests for every method and every usecase. Then we can be 100% sure about next releases.

PR #365, opened 2023-09-01
Fix "socket hang up" / ECONNRESET on consecutive requests with Node.js 19 and Node.js 20

Enflow-io:
## Changes ##

`node-fetch` version updated to ^2.6.13

## Reason ##

The reason is [the bug](https://github.com/node-fetch/node-fetch/issues/1735) when we cant send more than 1 request using Casper JS SDK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@types/lodash": "^4.14.191",
     "@types/mocha": "^5.2.7",
     "@types/node": "^14.14.31",
-    "@types/node-fetch": "^2.6.4",
+    "@types/node-fetch": "^2.6.13",
     "@types/sinon": "^10.0.13",
     "@typescript-eslint/eslint-plugin": "^5.52.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
     "@types/lodash": "^4.14.191",
     "@types/mocha": "^5.2.7",
     "@types/node": "^14.14.31",
-    "@types/node-fetch": "^2.6.4",
```
hoffmannjan: you changed `@types/node-fetch` instead of `node-fetch`

hoffmannjan (changes requested):
Thanks for the PR! You just changed `@types/node-fetch` instead of `node-fetch` package.
If you could change this we will be able to merge the PR and release it as new version. 

Cheers!

package.json
Enflow-io: I've fixed it. Sorry its Friday)

hoffmannjan (approved):

PR #387, opened 2024-02-05
fix: rename lib.js to lib.web.js

### LINKED ISSUE #386: Library cannot be used with modern moduleResolution TypeScript settings
## Description

It is not possible to import any type exported internally from the `lib` directory when your TS project is set to any of the modern values of `moduleResolution`: `"Node16"`, `"NodeNext"` or `"Bundler"`

### Explanation

The reason for that is filename clash between `./dist/lib.js` and `./dist/lib/` directory -- modern resolution algorithms resolve the `'./lib'` in
```ts
export * from './lib'
```
 in the `./dist/index.d.ts` file to the **`./dist/lib.js`** , not `./dist/lib/index.d.ts`, because that's how the Node resolution algorithm works (`lib.js` has a higher precedence than `lib/index.js`).

The quickest way to resolve the issue is to change the `lib.js` filename to something like `lib.web.js`, so that there is no confusion between the file name and the directory name.

### Prerequisites

- A TS Project  with a compilerOption `"moduleResolution"` set to `"Node16"`, `"NodeNext"` or `"Bundler"`

### Steps to reproduce

1. Try to import things like

```ts
import { CasperClient, CLPublicKey, DeployUtil } from 'casper-js-sdk';
```

### Expected behavior
The import is correct.

### Actual behavior
TypeScript error 2305:
```
Module '"casper-js-sdk"' has no exported member 'DeployUtil'
```

## Screenshots

Screenshot 2024-02-05 at 11 15 30

Screenshot 2024-02-05 at 11 15 35

## Additional context
The bug does not have any runtime implications, but it's not possible to import the library correctly in projects relying on current TypeScript defaults.

## Defect Type
- [ ] Regression
- [x] New
- [ ] Escape

akwodkiewicz:
This fixes TypeScript support for NodeNext/Node16/Bundler moduleResolution.

The issue was that the `./dist/lib.js` has a higher precedence than `./dist/lib/index.d.ts` when TypeScript tries to resolve `"./lib"` imports/exports.

Instead of renaming the whole `lib/` folder I decided to add a infix to the `lib.js` artifact. But to solve the issue you could've renamed the directory instead.

Resolves #386

### DIFF
--- a/.github/workflows/publish-casper-client-sdk.yml
+++ b/.github/workflows/publish-casper-client-sdk.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Get latest release version number
       id: get_version
       uses: battila7/get-version-action@d97fbc34ceb64d1f5d95f4dfd6dce33521ccccf5 #v2.3.0
-    - run: cp dist/lib.js casper-js-sdk.v${{ steps.get_version.outputs.version }}.js
+    - run: cp dist/lib.web.js casper-js-sdk.v${{ steps.get_version.outputs.version }}.js
     - uses: meeDamian/github-release@7ae19492500104f636b3fee4d8103af0fed36c8e #v2.0.3
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/casper-ecosystem/casper-js-sdk.git"
   },
   "main": "dist/lib.node.js",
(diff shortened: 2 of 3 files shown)


gyroflaw (approved):
Looks great, thanks for contribution

hoffmannjan (approved):
LGTM 👍

hoffmannjan:
Will be included in `2.15.4`.

PR #403, opened 2024-04-04
Feat 3.0 scratch

### LINKED ISSUE #412: Create project structure for JS SDK v3

hoffmannjan:
Closes https://github.com/casper-ecosystem/casper-js-sdk/issues/412


.github/workflows/ci-casper-client-sdk.yml
```diff
@@ -47,52 +47,64 @@ jobs:
         uses: actions/setup-node@5b949b50c3461bbcd5a540b150c368278160234a #v3.4.0
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'npm'
+          cache: 'yarn'
+
+        # https://github.com/dtinth/setup-github-actions-caching-for-turbo
+      - name: Setup Turbo cache server
+        uses: dtinth/setup-github-actions-caching-for-turbo@e26dc6b0473cc072107f6cb4f6e69d33ac2fc386 # v1.1.0
```
zajko, line 54: Do we have to rely on a commit dependency? It seems that there is a 1.2.0 tag in the repo for this plugin?

.github/workflows/ci-casper-client-sdk.yml
```diff
@@ -47,52 +47,64 @@ jobs:
...
         run: npm run docs:build
 
       - name: Test build
         run: npm run build
 
-      # test mainnet
-      - name: Get the mainnet version
-        id: get-mainnet-version
-        run: echo "MAINNET_VERSION=$(curl -s -X POST --data-raw '${{ env.VERSION_QUERY }}' -H "Content-Type:application/json" ${{ env.MAINNET_NODE_URL }} | jq -r .result.build_version | cut -d "-" -f 1)" >> "$GITHUB_ENV"
+      # # test mainnet
```
zajko: <This ius just a nitpick, ignore if you don't agree>
I know that functionally it is the same thing, but could we, instead of commenting these steps out use an if statement that is false?
```
- name: Get the mainnet version
   if ${{ false }}     
```
This should disable execution of the step, but it will still be interpreted so any syntax errors will be caught and we don't have to commit commented code. Also, after the if-statement we could have a comment explaining when this should be re-enabled.

.github/workflows/ci-casper-client-sdk.yml
ryoknz: we should use commit in the GitHub workflow, this is non-written rule.
https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

.github/workflows/ci-casper-client-sdk.yml
ryoknz: I agree with you

packages/keys/assets/ed25519/private_key.pem
```diff
@@ -0,0 +1,3 @@
+-----BEGIN PRIVATE KEY-----
```
zajko, line 1: These are test assets, right? I'm wondering if we could put them in some separate place. I would expect that folder `<sub-project>/assets` contains files that are needed for the actual compilation and runtime of the project.
This can be done in a later PR though.

.github/workflows/ci-casper-client-sdk.yml
zajko: 👍

webpack.config.karma.js
```diff
@@ -15,5 +15,12 @@ module.exports = {
       filename: null,
```
zajko: Turbo will read and correctly interpret this karma config?
Also, I see that we still have `webpack.config.js` -> should we still be keeping it? I see that it also is still listed as dependency

hoffmannjan: Same from my side

hoffmannjan: ^ Shouldn't we remove webpack 100%?

zajko (approved):
LGTM :+1:

codecov.yml
```diff
@@ -0,0 +1,33 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 80%
+        threshold: 1%
+    patch:
+      default:
+        target: auto
+        threshold: 5%
```
hoffmannjan, line 10: we will make it higher value over time ;)

PR #415, opened 2024-04-11
feat: add info_get_peers rpc as poc

### LINKED ISSUE #416: create a sub package for api layer

ryoknz:
Closes https://github.com/casper-ecosystem/casper-js-sdk/issues/416

## TODO

- [ ] Check if this [section ](https://github.com/casper-ecosystem/casper-js-sdk/pull/415/files#diff-eb64ebd2b79fb91f34e442c52d09308b15c35b8053ba17bfcc1fab2bd69c9097R13-R22) should be moved to `casper-js-types` module.


packages/api/src/CasperJsonRpc/informational/info_get_peers.ts
```diff
@@ -0,0 +1,60 @@
...
+  [ReturnType.Parsed]: GetPeersResult;
+};
+
+/**
+ * Returns a list of peers connected to the node
+ * @param baseJsonRPC
+ * @param options
+ * @returns
+ */
+export async function infoGetPeers<T extends keyof InfoGetPeersReturnTypeMap>(
```
zajko, line 50: Is this the function that should also handle calls like
```
infoGetPeers(__rpcUrl__, configuration) 
```
? Or will this be handled in another PR?

ryoknz: discussed this with Jan and he agreed to change the syntax like current implementation

PR #429, opened 2024-05-10
Stored value alignments pt.1

zajko:


src/services/CasperServiceByJsonRPC.ts
```diff
@@ -279,10 +297,10 @@ export class CasperServiceByJsonRPC {
...
         params: blockHash
           ? [
-              {
-                Hash: blockHash
-              }
-            ]
+            {
+              Hash: blockHash
+            }
+          ]
```
ryoknz: NIT: It seems you don't setup prettier stuff, these changes are not required, right?

ryoknz (approved):
Great job, 👍

src/services/CasperServiceByJsonRPC.ts
zajko: :+1:

PR #430, opened 2024-05-13
Stored value alignments pt2

zajko:


src/lib/EntryPointAccess.ts
```diff
@@ -0,0 +1,43 @@
+import { TypedJSON, jsonArrayMember, jsonObject } from "typedjson";
+
+export abstract class EntryPointAccess {
+    public abstract toJSON(): unknown
+}
+
+export class PublicEntryPointAccess extends EntryPointAccess {
+    public toJSON(): unknown {
```
ryoknz:
```suggestion
    public toJSON(): string {
```

src/lib/EntryPointAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+        return "Public"
+    }
+}
+
+export class TemplateEntryPointAccess extends EntryPointAccess {
+    public toJSON(): unknown {
```
ryoknz:
```suggestion
    public toJSON(): string {
```

src/lib/EntryPointAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+        return "Template"
+    }
+}
+
+@jsonObject
+export class GroupsEntryPointAccess extends EntryPointAccess {
+    @jsonArrayMember(String)
+    public Groups: string[];
+
+    public toJSON(): unknown {
```
ryoknz:
```suggestion
    public toJSON() {
```

ryoknz (approved):
LGTM 👍 , we should test the new data types with devnet in the target branch,

src/lib/EntryPointAccess.ts
zajko: It needs to be unknown (or any) since 'GroupsEntryPointAccess' returns an object

src/lib/EntryPointAccess.ts
zajko: As above

src/lib/EntryPointAccess.ts
ryoknz: I think its okay `string` for PUblicEntryPointAccess, TemplateEntryPointAccess , in SDK, we should avoid using unkown or any as possible, but we can do this later.

src/lib/EntryPointAccess.ts
zajko: :+1:

PR #434, opened 2024-05-21
Fixing e2e pt1

zajko:


src/lib/AccountIdentifier.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(key: string) {
+    this.key = key;
+  }
+
+  toJSON(): string {
+    return this.key;
+  }
+}
+
+export class AccountHash implements AccountIdentifier {
```
ryoknz: Why don't we use existing classes for these?

zajko: Could you point me to the class that is a polymorphic representation of either PublicKey or AccountHash? I didn't find it.

ryoknz: https://github.com/casper-ecosystem/casper-js-sdk/blob/b6b6889e75605b28d10eb1c390435b1f582b9e87/src/lib/CLValue/PublicKey.ts#L72

https://github.com/casper-ecosystem/casper-js-sdk/blob/b6b6889e75605b28d10eb1c390435b1f582b9e87/src/lib/CLValue/AccountHash.ts#L46

zajko: The `state_get_account_info` RPC method doesn't take a `CLValue`. It takes an `AccountIdentifier`.
So we still need something that knows how to make a hex-string like "017b14afdcaa0c2890bc317352aff880e3d5d4d703500e4d2415de93daaa43b64e" from a CLPublicKey or "account-hash-73911a773cd124ae43572066dedb321e71d93dd18c337ef1365c8bc19e95895f" from CLAccountHash

zajko: Ok, after some thought I just changed the type of accountIdentifier to CLPublicKey, since it has `toAccountHashStr`. Our API clients don't have to implement every possible alternative that arguments can be constructed to the json API

ryoknz: Thanks for updates, but please use `CLAccountHash` for identifier instead of `CLPublicKey`.

hoffmannjan (approved):

Issue #438, opened 2024-06-10
Add tests for CLKey variants


hoffmannjan:
based on example JSONs test if parsers works correctly

PR #439, opened 2024-06-10
More supported variants of CLKey

hoffmannjan:
missing tests specified in [this ticket](https://github.com/casper-ecosystem/casper-js-sdk/issues/438)


src/lib/CLValue/KeyVariants.ts
```diff
@@ -658,10 +567,91 @@ export class BidAddr implements CLKeyVariant {
...
+
+export class BidAddr implements CLKeyVariant {
+  keyVariant = KeyTag.BidAddr;
+  prefix = BID_ADDR_PREFIX;
+
+  bidAddrTag = BidAddrTag;
+
+  constructor(public data: BidAddrData) {}
+
+  value(): any {
```
ryoknz:
```suggestion
  value(): BidAddrData {
```

can we have type here?

src/lib/CLValue/KeyVariants.ts
```diff
@@ -684,3 +674,33 @@ export class BidAddr implements CLKeyVariant {
...
   }
 }
+
+export class KeyPackage implements CLKeyVariant {
+  keyVariant = KeyTag.Package;
+  prefix = PACKAGE_PREFIX;
+
+  constructor(public data: Uint8Array) {}
+
+  value(): any {
```
ryoknz:
```suggestion
  value(): Uint8Array {
```

same here

ryoknz (approved):
Looks great, left some NITs
and please take a look failed test, https://github.com/casper-ecosystem/casper-js-sdk/actions/runs/9641365736/job/26586883018#step:8:19

src/lib/CLValue/Key.ts
```diff
@@ -19,15 +15,47 @@ import {
...
+  BID_PREFIX,
+  WITHDRAW_PREFIX,
+  DICTIONARY_PREFIX,
+  SYSTEM_ENTITY_REGISTRY_PREFIX,
+  ERA_SUMMARY_PREFIX,
+  UNBOND_PREFIX,
+  CHAINSPEC_REGISTRY_PREFIX,
+  CHECKSUM_REGISTRY_PREFIX,
+  BidAddr,
+  BidAddrParser
```
zajko, line 50: Do we allow asterisk imports? Maybe it would be cleaner to just do `import * from './index'`?

zajko (approved):

src/lib/CLValue/Key.ts
hoffmannjan: hm, do we use asterisk imports anywhere? its good idea but I would add this in 4.0.0 as standard

PR #444, opened 2024-06-12
Fixed changelog for rc00, some minor codefixes, reorganizations of co…

zajko:
…de. Reverted some namechanges to make the changediff less of an eyesore


ryoknz (approved):

CHANGELOG.md
```diff
@@ -15,9 +15,45 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  - `Message`
+  - `Reservation`
+  - `EntryPoint`
+- `TransactionUtil` class with all the subclasses to support sending Transaction to the node
+- `transaction`, `getAccountInfo`, `queryBalanceDetails`, `getTransactionInfo`, `waitForTransaction`, `queryGlobalState` methods added to `CasperServiceByJsonRPC` class
+- `Bid` class has new fields: `validator_public_key`, `vesting_schedule`
+- `merkleProof` field to `EraSummary` class
+- optional `latest_switch_block_hash` added to `GetStatusResult`
+
+## Breaking changes compared to 2.x
```
hoffmannjan: I know this is special situation but I would follow the https://keepachangelog.com/en/1.1.0/ convention so this probably should be just `## Changed` and then in description we can add that its compared to previous stable version which was 2.x. 

I would do this because if we will want to generate some documents from this CHANGELOG file it should make it easier.

hoffmannjan (approved):
One nitpick - overall awesome work and it nice to see what was changed over this version preparation. Thank you @zajko !

hoffmannjan (approved):

PR #449, opened 2024-07-18
Making it possible to specify contract by hash and name.

zajko:
Removing `TransactionSessionKind` since field `kind` was removed from `Session` implementation of `TransactionTarget`


jnhfmn (approved):
LGTM

hoffmannjan (approved):

PR #482, opened 2024-12-23
Bugfix: entrypoint serialization

davidatwhiletrue:
String was serialized as a calltable while it must be a length-prefixed byte array.


alexmyshchyshyn (approved):

alexmyshchyshyn:
LGTM

Issue #483, opened 2024-12-31
newCLMap - Doesn't return correct object


KillianH:
https://github.com/casper-ecosystem/casper-js-sdk/blob/feat-5.0.0/src/types/clvalue/Map.ts#L235

It return a new and empty CLValue instance instead of returning the CLValue created at L233

PR #484, opened 2025-01-02
Fix: CLValueMap deserialization from bytes, added unit tests for Tuple, Map, Numeric CLValues

alexmyshchyshyn:
- <ins>Resolves</ins>: https://github.com/casper-ecosystem/casper-js-sdk/issues/483
- Fixed `CLValueMap` deserialization from bytes
- Added unit tests


davidatwhiletrue (approved):
LGTM

PR #497, opened 2025-01-28
Fix issue with BlockTransaction serialisation and rewardedSignatures …

Comp0te:
…type


src/types/Block.ts
```diff
@@ -895,18 +905,17 @@ export class BlockBodyV2 {
   /**
    * The list of transactions included in the block.
    */
-  @jsonArrayMember(BlockTransaction, {
+  @jsonMember(BlockTransaction, {
```
alexmyshchyshyn, line 915: Why was `jsonArrayMember` replaced with `jsonMember` if transactions are an array?

Comp0te: Because in json itself, this is not an array, but an object

alexmyshchyshyn (approved):

PR #504, opened 2025-02-03
 Add new get auction info v2 method support

alexmyshchyshyn:


src/rpc/rpc_client.ts
```diff
@@ -784,6 +823,31 @@ export class RpcClient implements IClient {
...
+    try {
+      const resV2 = await this.getAuctionInfoV2ByHash(blockHash);
+      const result = new StateGetAuctionInfoResult();
+      result.auctionState = AuctionState.fromV2(resV2.auctionState);
+      result.version = resV2.version;
+      result.rawJSON = resV2.rawJSON;
+      return result;
+    } catch (err) {
+      const errorMessage = err?.message || '';
+      if (!errorMessage.includes('Method not found')) {
```
davidatwhiletrue, line 835: @alexmyshchyshyn  getAuctionInfoV2ByHash can fail with next error for a block generated in Casper 1.x:

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32003,"message":"Query failed","data":"value was not found in the global state"}}
```

the flow should continue in this case to use v1 method, right?

alexmyshchyshyn: The devnet has not been switched to the latest RC version. Therefore, the error from `getAuctionInfoV2ByHash`—specifically, `"Query failed"` for blocks generated in Casper 1.x—should not occur in new environment.

Comp0te (approved):

PR #531, opened 2025-03-13
Add method to convert Transaction to Deploy

alexmyshchyshyn:


src/types/Deploy.ts
```diff
@@ -443,6 +443,27 @@ export class Deploy {
     );
   }
 
+  public static newDeployFromTransaction(tx: Transaction): Deploy {
```
davidatwhiletrue, line 446: @alexmyshchyshyn can tx be already a Deploy?

src/types/ExecutableDeployItem.ts
```diff
@@ -614,4 +615,69 @@ export class ExecutableDeployItem {
...
+   */
+  public static newExecutableDeployItemFromTransaction(
+    transaction: Transaction
+  ): ExecutableDeployItem {
+    const executableDeployItem = new ExecutableDeployItem();
+    const entryPoint = transaction.entryPoint.customEntryPoint ?? '';
+    const { target, args } = transaction;
+
+    if (target?.native !== undefined) {
+      executableDeployItem.transfer = new TransferDeployItem(args);
```
davidatwhiletrue, line 633: should this fail if the native transaction is not a transfer but an auction related operation?

davidatwhiletrue (commented):
@alexmyshchyshyn check my comments.

src/types/Deploy.ts
alexmyshchyshyn: Yes, will add a condition to check this. Thanks

src/types/ExecutableDeployItem.ts
alexmyshchyshyn: Good catch @davidatwhiletrue . 
So, if the transaction target is an auction-related operation, it will have native. 
I propose to throw an error in this case, cause we can not build auction deploys from the transaction.

alexmyshchyshyn:
Per discussion, we took a decision to avoid it, cause Transaction can not be converted to Deploy properly.

PR #535, opened 2025-03-17
Add methods to wait for transaction / deploy

alexmyshchyshyn:


src/rpc/rpc_client.ts
```diff
@@ -1381,6 +1382,72 @@ export class RpcClient implements IClient {
...
+    while (true) {
+      const info = await getInfo(hash);
+      const executionResult = (info as any)?.executionInfo?.executionResult;
+
+      let successful = false;
+
+      if (!executionResult) {
+        successful = false;
+      } else {
+        return info;
```
davidatwhiletrue: this returns without clearing the timer.

src/rpc/rpc_client.ts
```diff
@@ -1381,6 +1382,72 @@ export class RpcClient implements IClient {
...
+      }
+
+      if (successful) {
```
davidatwhiletrue: `successful` is never `true`. Do we need this var?
Maybe:
```
    while (true) {
        const transactionInfo = await rpcClient.getTransactionByDeployHash(deployHash); // YOu will need to create RPC client

        const execution_result = transactionInfo.executionInfo?.executionResult
        if (execution_result) {
            clearTimeout(timer);
            return transactionInfo;
        } else {
            await sleep(400);
        }
    }
```

davidatwhiletrue (changes requested):

src/rpc/rpc_client.ts
alexmyshchyshyn: Refactored function, also handled scenarios, when requests could fail

Comp0te (approved):

PR #536, opened 2025-03-19
Provide possibility to specify the gas price tolerance when creating Transactions and Deploys

alexmyshchyshyn:


davidatwhiletrue (approved):
LGTM

PR #549, opened 2025-04-03
Change hash type to TransactionHash

alexmyshchyshyn:


src/types/Transaction.ts
```diff
@@ -63,6 +62,111 @@ export enum TransactionVersion {
...
+  public transactionV1?: Hash;
+
+  /**
+   * Constructs a new `TransactionHash` instance, which can hold either a `Deploy` hash or a `TransactionV1` hash.
+   * @param deploy The hash of the deploy transaction, if applicable.
+   * @param transactionV1 The hash of the version 1 transaction, if applicable.
+   */
+  constructor(deploy?: Hash, transactionV1?: Hash) {
+    this.deploy = deploy;
+    this.transactionV1 = transactionV1;
```
ihor: Maybe let's make the constructor private and add two factory methods:
- `fromDeployHash`
- `fromTransactionHash`

src/types/key/Hash.ts
```diff
@@ -1,12 +1,25 @@
...
+  equals(other: Hash): boolean;
+}
+
 /**
  * Represents a cryptographic hash.
  * This class provides methods for creating, manipulating, and comparing hash values.
  */
 @jsonObject
-export class Hash {
+export class Hash implements IHash {
```
ihor: This change would require the consumers to switch to the interface where they relied on the `Hash` type. Maybe let's inherit `TransactionHash` from `Hash` instead?

ihor (changes requested):
@alexmyshchyshyn please check my comments

Comp0te (approved):

ihor (approved):

PR #559, opened 2025-04-22
Add query latest balance utility function for Casper Network

alexmyshchyshyn:


src/utils/casper-network.ts
```diff
@@ -349,4 +351,31 @@ export class CasperNetwork {
...
+      try {
+        const stateBalanceResult = await this.rpcClient.getLatestBalance(
+          identifier.purseUref.toPrefixedString()
+        );
+
+        if (stateBalanceResult) {
+          balanceResult.balance = stateBalanceResult.balanceValue;
+          balanceResult.apiVersion = stateBalanceResult.apiVersion;
+        }
+      } catch (error) {
```
ihor: This try/catch section seems to be redundant

Comp0te (approved):

ihor (approved):

PR #571, opened 2025-06-13
Disable nightly scheduled pipeline

alexmyshchyshyn:


Comp0te (approved):
We disable this workflow for now.
After e2e is finished -> we will reenable it

PR #573, opened 2025-08-01
add protocol_major_version as optional property in ByPackageHash invo…

alexmyshchyshyn:
…cations

### DIFF
--- a/src/types/TransactionTarget.ts
+++ b/src/types/TransactionTarget.ts
@@ -9,7 +9,8 @@ import {
   CLTypeUInt32,
   CLValue,
   CLValueOption,
-  CLValueString
+  CLValueString,
+  CLValueUInt32
 } from './clvalue';
 import { ExecutableDeployItem } from './ExecutableDeployItem';
 import { CalltableSerialization } from './CalltableSerialization';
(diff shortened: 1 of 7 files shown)


src/types/TransactionTarget.ts
```diff
@@ -135,6 +135,14 @@ export class ByPackageHashInvocationTarget {
   @jsonMember({ name: 'version', isRequired: false, constructor: Number })
   version?: number;
 
+  @jsonMember({
+    name: 'protocol_version_major',
+    constructor: Number,
+    isRequired: false,
+    preserveNull: false
+  })
+  public protocolVersionMajor?: number | null;
```
davidatwhiletrue: I'd try to avoid undefined here. Maybe initializing to null in the constructor.

src/types/TransactionTarget.ts
```diff
@@ -333,6 +351,7 @@ export class TransactionInvocationTarget {
         const byPackageHash = new ByPackageHashInvocationTarget();
         byPackageHash.addr = packageHash.result;
         byPackageHash.version = BigNumber.from(version).toNumber();
+        byPackageHash.protocolVersionMajor = null;
```
davidatwhiletrue: we can have a field `3` in the calltable with a value, right?

davidatwhiletrue (commented):
protocolVersionMajor is missing in ByPackageName variant.

Issue #574, opened 2025-09-01
gas_price not set to value indicated in payment-gasPriceTolerance


davidatwhiletrue:
**Describe the bug**

When building a Deploy using the `TransactionBuilder<T>`, it is expected that the `gasPriceTolerance` indicated in `payment()` is used for the `gas_price` value in the Deploy header. But `gasPrice=1` always.

```javascript
    const builder = new SessionBuilder()
        .from(global.newAccount300Key.publicKey)
        .installOrUpgrade()
        .runtimeArgs(args)
        .wasm(new Uint8Array(contractWasm))
        .payment(paymentAmount, 2) 
        .chainName(NETWORK_NAME)
        .buildFor1_5();
```

**To Reproduce**

Identify if any Prequisites:

2.  ..

**Expected behavior**

**Screenshots**

**Additional context**

**Defect Type**
- [ ] Regression
- [ ] New
- [ ] Escape

PR #575, opened 2025-09-01
fix an issue when gas_price not set to value indicated in payment-gasPriceTolerance

alexmyshchyshyn:
- <ins>Resolves</ins>: https://github.com/casper-ecosystem/casper-js-sdk/issues/574


davidatwhiletrue (approved):
LGTM

Issue #595, opened 2026-04-01
Node.js v22 Deprecation: url.parse() via @open-rpc/client-js


tomasoliveirz:
Hi team!

Using `casper-js-sdk` in Node.js v22 triggers the following deprecation warning:

```(node:70238) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.```

The warning surfaces through the @open-rpc/client-js dependency `node_modules/@open-rpc/client-js/src/__mocks__/requestData.ts`

While it is currently just a console warning, the legacy URL API has known parsing inconsistencies that could lead to security issues. 

Could we update the dependency tree or migrate to the WHATWG URL API to resolve this and ensure better security?

**Environment:** 
- Node.js: v22.22.2
- casper-js-sdk: v5.0.6

Thanks!

alexmyshchyshyn:
Hi @tomasoliveirz,

Thank you for reporting the issue! 

These fixes have been implemented and will be available in the upcoming release: **v5.0.11**.

If you have any further questions or feedback, feel free to reach out. Thanks for helping us improve! 🚀

Issue #567, opened 2025-05-09
Imports not working in modern TypeScript project


1337jay:
Details:

Imports are not working in a modern TypeScript project

Reporduce:

"compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "outDir": "./dist",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "noImplicitAny": true,
    "skipLibCheck": true,
    "moduleResolution": "Node10",
    "resolveJsonModule": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "types": ["@wdio/globals/types"]
  },

import {
  RpcClient,
  HttpHandler,
  PublicKey,
  PrivateKey,
  KeyAlgorithm,
  Deploy,
  DeployHeader,
  ExecutableDeployItem,
  TransferDeployItem
} from 'casper-js-sdk';

**Expected behavior**
Imports work.

**Additional context**
SyntaxError: The requested module 'casper-js-sdk' does not provide an export named 'Deploy'

**Defect Type**
- [x] Regression
- [ ] New
- [ ] Escape

alexmyshchyshyn:
Hi @1337jay,

Thank you for reporting the issue! 
These fixes have been implemented and will be available in the upcoming release: **v5.0.11**.

If you have any further questions or feedback, feel free to reach out. Thanks for helping us improve! 🚀

PR #596, opened 2026-04-01
Resolve DEP0169 url.parse() deprecation and add ESM support

alexmyshchyshyn:
- **Removed unused `@open-rpc/client-js` dependency**
  - Its compiled mock invoked `url.parse()`, triggering `DEP0169` warnings during test runs  

- **Switched Axios to the built-in `fetch` adapter (`adapter: 'fetch'`)**
  - Bypasses `follow-redirects`, which was calling `url.parse()` on every RPC request  

- **Fixed `HttpError` being swallowed in `processFetchRequest`**
  - Moved the status check outside the `try/catch` so errors propagate correctly to callers  

- **Added Webpack ESM build target (`lib.esm.js`)**
  - Uses `experiments.outputModule`  

- **Updated `package.json`**
  - Added `"exports"` and `"module"` fields for proper ESM/CJS resolution  
  - Fixes named imports in modern TypeScript projects  

- **Added `"engines": { "node": ">=18" }`**
  - Reflects the requirement for the `fetch` adapter  

- **Added unit tests for `HttpHandler`**
  - Covers both Axios and fetch client paths  

- <ins>Resolves</ins>: https://github.com/casper-ecosystem/casper-js-sdk/issues/595
- <ins>Resolves</ins>: https://github.com/casper-ecosystem/casper-js-sdk/issues/567

### DIFF
--- a/.github/workflows/ci-casper-client-sdk.yml
+++ b/.github/workflows/ci-casper-client-sdk.yml
@@ -61,7 +61,7 @@ jobs:
         run: npm run lint:ci
 
       - name: Unit Test
-        run: npm run test:node:unit
+        run: ${{ matrix.node-version == '22.x' && 'npm run test:node:unit:node22' || 'npm run test:node:unit' }}
 
       - name: Test build
         run: npm run build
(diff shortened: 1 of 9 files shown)


davidatwhiletrue (approved):
LGTM

PR #603, opened 2026-08-06
Fix production npm audit advisories and restrict published package contents (5.0.13)

Comp0te:
Restores the CI production audit gate — `npm audit --omit=dev` is currently failing on `dev`, so the **Audits** step is red — and tightens what the npm tarball ships. Released as `5.0.13`.

No source files change. Everything here is dependency, lockfile and packaging metadata.

## Dependency fixes

| Package | Change | Why |
|---|---|---|
| `axios` | `^1.15.0` → `^1.19.0` | Clears 18 advisories (ReDoS via cookie name injection, unbounded allocation, prototype-pollution gadgets). The advisory range closes below `1.18.0`. Also hard-floors `form-data` at the fixed `4.0.6` |
| `brace-expansion` (override) | `^5.0.5` → `^5.0.9` | Clears four DoS advisories. The old override was itself the problem — it floored at the vulnerable `5.0.5`. `5.0.9` is the *minimum* viable floor: the newest advisory, `GHSA-rgw5-rvv9-x895`, is fixed only there |
| `form-data` | → `4.0.6` | CRLF injection fix, now enforced by axios rather than incidental |
| `glob` | removed | Dead dependency — zero usage anywhere in the repo |

Both floors had to be raised rather than relying on `npm install`: every fix was already published *within* the declared ranges, but npm never replaces a locked version that still satisfies its range, so `axios@1.15.2` and `brace-expansion@5.0.5` stayed pinned. The lockfile is fully regenerated.

### Why `glob` had to go rather than be upgraded

`glob` was the only path to `inflight@1.0.6`, which carries a memory-leak advisory (`sonatype-2023-4801`, CVSS 6.1). That advisory has **no GHSA entry**, so `npm audit` cannot report it, and **no fixed version exists** — the package is abandoned and deprecated by its own author. Removing the dead dependency is the only available remedy.

Verified unused six ways before removal: no `import`/`require`/dynamic `import()` anywhere in the repo; the literal string `glob` appears only in its own `package.json` entry (which also rules out a computed `require`); not used by any npm script; not referenced by any webpack/karma/tsconfig/eslint config; absent from both built bundles; and with it removed, lint, the unit suite and the build all pass on Node 18 and 22.

Removing it also drops `minimatch`, `brace-expansion`, `balanced-match`, `once`, `wrappy`, `fs.realpath` and `path-is-absolute` from the production tree, taking it from 59 packages to 54. The `minimatch` and `brace-expansion` overrides stay — the dev tree still needs them.

## Packaging

`package.json` had no `files` field, so npm shipped everything the `.npmignore` denylist did not explicitly name: the vocs docs site, CI workflows, editor and tooling config, and the repo's own build configs. Replaced with a whitelist, so anything added to the repo later stays out of the tarball unless deliberately listed:

```json
"files": ["dist", "resources", "README.md", "CHANGELOG.md", "LICENSE"]
```

Internal working documents move to `internal-docs/`, now excluded in both `.gitignore` and `.npmignore`.

### Published package contents

|  | 5.0.12 | 5.0.13 | Delta |
|---|---:|---:|---:|
| Files | 341 | **163** | **−178 (−52%)** |
| Unpacked | 8.09 MB | **7.86 MB** | −0.23 MB (−2.8%) |
| Tarball (gzip) | 1.76 MB | **1.62 MB** | −0.14 MB (−8.0%) |

Measured against the actually published `casper-js-sdk@5.0.12` tarball pulled from the registry, not an estimate. The byte saving is far smaller than the file-count drop because two effects cancel: the whitelist removes −0.43 MB, while `dist/` grows +0.20 MB from the rebuild with axios 1.19. `dist/` is ~99% of the payload, and roughly 5 MB of it is source maps.

`site/` is the bulk of the removed files (163). It is a vocs app that cannot function from inside `node_modules` — `vocs` is a devDependency, so consumers never receive it — and the docs workflow builds it from a repo checkout, never from the package. Deep imports into it were already blocked by the `exports` field added in 5.0.12.

## Verification

`npm audit --omit=dev` reports 0 vulnerabilities, and `lint:ci`, the unit suite (170 passing) and `build` are green on **both Node 18.20.4 and 22.22.2**, mirroring the CI matrix.

The packed tarball was additionally installed into a clean fixture and exercised: 438 symbols export, `main`/`browser`/`types` and all five `exports` conditions resolve to files that exist, and `resources/` plus the metadata files are present while `site/`, `.github/`, config files and `internal-docs/` are absent.

## Note for reviewers

`npm run test:browser:unit` fails, and does so identically on `dev` before this branch — it is not caused by these changes and is not run by CI. Two independent pre-existing causes: `karma@6.4.4` calls `require('minimatch')` as a function while the repo's `minimatch ^10.2.3` override supplies v10, which has no callable CJS default; and behind that, `karma.conf.js` preprocesses `src/{rpc,sse,types}/**/*.ts` while all 30 test files now live under `src/tests/**`, so the run reports `Executed 0 of 0`. Both are resolved by the planned Vitest migration; tracked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,18 +17,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Changed
 
-- `axios` bumped to `^1.19.0`. The previous `^1.15.0` floor let the lockfile keep `1.15.2`, which carried 18 advisories — ReDoS via cookie name injection, unbounded allocation, and several prototype-pollution gadgets. The advisory range closes below `1.18.0`. No source change was required: the SDK's use of `axios.create({ adapter: 'fetch' })`, `request()` and `isAxiosError()` is stable across the bump
-- `brace-expansion` override raised from `^5.0.5` to `^5.0.9`, clearing four DoS advisories. The old override was itself the problem — it pinned the vulnerable `5.0.5` as its floor
-- `form-data` now resolves to `4.0.6`, which fixes a CRLF injection advisory. `axios@1.19.0` requires it, so the floor is enforced rather than incidental
-- `package-lock.json` fully regenerated. Every fix above was already published within the declared ranges, but `npm install` never replaces a locked version that still satisfies its range
+- Restricted the published package to `dist`, `resources`, `README.md`, `CHANGELOG.md` and `LICENSE` via a `files` whitelist — the docs site, CI workflows and build configs no longer ship
+- Raised `axios` to `^1.19.0`, the `brace-expansion` override to `^5.0.9` and `form-data` to `4.0.6`, and fully regenerated `package-lock.json`. No source change was required
 
 ### Removed
 
-- `glob` production dependency. It had zero usage anywhere in the repo and was the only path to `inflight@1.0.6`, an abandoned package with a memory-leak advisory (`sonatype-2023-4801`) that has no fixed version and no GHSA entry — meaning `npm audit` could not report it. Removing `glob` also drops `inflight`, `minimatch`, `brace-expansion`, `balanced-match`, `once`, `wrappy`, `fs.realpath` and `path-is-absolute` from the production tree, taking it from 59 packages to 54
+- Unused `glob` production dependency, and with it the abandoned `inflight` and the rest of its transitive tree
 
 ### Fixed
 
-- All production-reachable dependency advisories: `npm audit --omit=dev` now reports 0 vulnerabilities, restoring the CI audit gate
+- Security vulnerabilities in production dependencies
 
 ### [5.0.12] - 2026-04-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
   ### Removed
  -->
 
+### [5.0.13] - 2026-08-06
+
+### Changed
+
+- `axios` bumped to `^1.19.0`. The previous `^1.15.0` floor let the lockfile keep `1.15.2`, which carried 18 advisories — ReDoS via cookie name injection, unbounded allocation, and several prototype-pollution gadgets. The advisory range closes below `1.18.0`. No source change was required: the SDK's use of `axios.create({ adapter: 'fetch' })`, `request()` and `isAxiosError()` is stable across the bump
+- `brace-expansion` override raised from `^5.0.5` to `^5.0.9`, clearing four DoS advisories. The old override was itself the problem — it pinned the vulnerable `5.0.5` as its floor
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- `form-data` now resolves to `4.0.6`, which fixes a CRLF injection advisory. `axios@1.19.0` requires it, so the floor is enforced rather than incidental
+- `package-lock.json` fully regenerated. Every fix above was already published within the declared ranges, but `npm install` never replaces a locked version that still satisfies its range
+
+### Removed
+
+- `glob` production dependency. It had zero usage anywhere in the repo and was the only path to `inflight@1.0.6`, an abandoned package with a memory-leak advisory (`sonatype-2023-4801`) that has no fixed version and no GHSA entry — meaning `npm audit` could not report it. Removing `glob` also drops `inflight`, `minimatch`, `brace-expansion`, `balanced-match`, `once`, `wrappy`, `fs.realpath` and `path-is-absolute` from the production tree, taking it from 59 packages to 54
+
+### Fixed
+
+- All production-reachable dependency advisories: `npm audit --omit=dev` now reports 0 vulnerabilities, restoring the CI audit gate
```
davidatwhiletrue: this could be a lie by the time we push the release :-)
Do we really need it ?

CHANGELOG.md
```diff
@@ -13,6 +13,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
  -->
 
+### [5.0.13] - 2026-08-06
+
+### Changed
+
+- `axios` bumped to `^1.19.0`. The previous `^1.15.0` floor let the lockfile keep `1.15.2`, which carried 18 advisories — ReDoS via cookie name injection, unbounded allocation, and several prototype-pollution gadgets. The advisory range closes below `1.18.0`. No source change was required: the SDK's use of `axios.create({ adapter: 'fetch' })`, `request()` and `isAxiosError()` is stable across the bump
+- `brace-expansion` override raised from `^5.0.5` to `^5.0.9`, clearing four DoS advisories. The old override was itself the problem — it pinned the vulnerable `5.0.5` as its floor
+- `form-data` now resolves to `4.0.6`, which fixes a CRLF injection advisory. `axios@1.19.0` requires it, so the floor is enforced rather than incidental
+- `package-lock.json` fully regenerated. Every fix above was already published within the declared ranges, but `npm install` never replaces a locked version that still satisfies its range
```
davidatwhiletrue: Should we simplify this four items? Maybe merging them into one?

davidatwhiletrue (approved):
@Comp0te changes look good. See my comments about CHANGELOG. I wonder if it's too verbose.

PR #604, opened 2026-08-07
Land the knip dead-dependency gate, remove dead dependencies, apply safe bumps

Comp0te:
> **Stacked on #603.** Base is `CSDK/prod-audit-hotfix`, so the diff here is the single commit `09feca03` — the hotfix commits below it are reviewed in #603. GitHub retargets this PR to `dev` automatically once #603 merges.

Low-risk hygiene pass, PHASE-1 of the dependency modernization plan. **No source file under `src/` is added, deleted or modified** — the only non-`package.json` changes are `.eslintrc.js`, the new `knip.json`, and the CI workflow.

## knip lands first, as a blocking gate

knip is added and wired into CI as a blocking step for dependency-level findings (`npm run knip`; `npm run knip:all` gives the full report). It runs *first* so it produces the removal list rather than verifying one after the fact — and it covers the one class a repo-wide grep structurally cannot see: **phantom dependencies**, imported but undeclared and masked by npm hoisting. The baseline reported none.

`knip.json` carries a rule: every `ignoreDependencies` entry must name the mechanism by which the dependency is genuinely used and knip cannot see it. "knip is wrong" is not admissible — if no such mechanism exists, the dependency is dead and gets removed instead. The four current entries are all browser polyfills whose package names collide with Node builtins.

knip's unused-*file* and unused-*export* findings are captured as an inventory only. Deleting dead code needs a public-API tripwire this repo does not have yet, and in a library an export unused *inside* the repo may still be load-bearing for consumers.

## Production dependencies removed

Each was flagged by knip **and** independently confirmed by grep over `src/`, `e2e/` and the config files:

| Removed | Why it is safe |
| --- | --- |
| `node-fetch` | Zero imports. `src/rpc/http_handler.ts` uses the axios fetch adapter and native `global.fetch`, available on Node ≥18 and in the web bundle. |
| `@scure/bip32` | Zero usages; no HD-wallet code in the repo. |
| `@scure/bip39` | Zero usages; no mnemonic/seed-phrase code. |
| `reflect-metadata` | Never imported — see below. |

**`reflect-metadata` deserves the closest look**, because static unreachability alone would not justify removing it. The argument is that typedjson never takes the reflection path here: it guards every `Reflect.getMetadata` access behind `isReflectMetadataSupported`, and every `@jsonMember` in `src` declares its type explicitly (a positional constructor, `constructor:`, or a `serializer`/`deserializer` pair, which bypasses type-driven conversion entirely). `emitDecoratorMetadata: true` stays on but is inert: tslib's `__metadata` is a guarded no-op when `Reflect.metadata` is absent. typedjson does not declare it as a peer dependency, so no missing-peer warning either. Verified by running the typedjson-heavy tests (InitiatorAddr / StoredValue / Bid) with `Reflect.getMetadata === undefined`.

Consumers who were relying on the SDK to pull `reflect-metadata` into their own tree will now have to declare it themselves. That, plus the removals above, is why the release is a minor rather than a patch.

## devDependencies removed

`casper-node-launcher-js` (unreferenced by anything, and the sole root of a HIGH `@oclif` advisory), `chai-as-promised`, `serve`, `karma-coverage-istanbul-reporter` and `@jsdevtools/coverage-istanbul-loader` (both unwired), the three ESLint plugins whose rules were never enabled (`eslint-plugin-jsdoc`, `eslint-plugin-lodash`, `eslint-plugin-prefer-arrow`), and the types for the removed packages.

## Bumps and overrides

Patch/minor bumps within the same major, zero code change: `typedjson`, `ts-results`, `humanize-duration`, `bn.js`, `assert`, `url`, several `@types/*`, and de-risking interim bumps of the crypto/ethers libraries (still dual-CJS, no API change). These are the resting state for the whole 5.x line — the noble/ethers *replacements* are deferred.

`lodash` and `lodash-es` overrides clear the multi-rooted dev-only code-injection advisories that parent upgrades cannot fix on their own. The `serialize-javascript` half of that group is deferred to PHASE-2, where the build step is gated off the Node 18 leg — every patched release of it needs Node ≥19.

## CI matrix

Moves to `[18.x, 22.x, 24.x]`. Node 18 stays as the consumer-compatibility leg for as long as the published `engines.node` is `">=18"`; 24 is the Active LTS; 20 is skipped as EOL and adds no coverage the 18 floor-leg and the 22/24 legs do not already bound.

The knip step is gated off the 18.x leg because `knip@6` requires Node `^20.19 || >=22.12` — this makes PHASE-1, not PHASE-2, the first phase to introduce a Node-≥20.9 devDependency, so the step-level gating starts here. `npm ci` on Node 18 still succeeds; only execution is gated. The `node22` test variant now covers 24.x as well, since Node ≥23 also enables native TS type-stripping by default.

The published `engines.node` is unchanged.

## Verification

Verified on Node 18.20.4, 22.22.2 and 24.19.0: `npm audit --omit=dev` reports 0, `lint:ci`, the unit suite and `build` are green, and `npm run knip` exits 0 on the 22/24 legs.

`npm run test:browser:unit` is **known red and pre-existing** — it fails the same way on a clean install of the committed `dev` tree, before any change in this PR. `karma@6.4.4` calls `require('minimatch')` as a function while the repo's `minimatch ^10.2.3` override supplies v10, which has no callable CJS default; and past that it runs 0 of 0 tests, because `karma.conf.js` preprocesses `src/{rpc,sse,types}/**/*.ts` while all 30 test files live under `src/tests/**`. Resolved structurally in PHASE-3, which deletes karma outright.


davidatwhiletrue (commented):
Should you update version to `5.1.0` here so that this doesn't go accidentally into `5.0.13`?

Comp0te:
@davidatwhiletrue I bumped it to the next PR and will release it together with assigning the proper tag in the commit with a version bump

davidatwhiletrue (approved):
LGTM

PR #605, opened 2026-08-07
Modernize the dev toolchain, isolate the docs site, release 5.1.0

Comp0te:
> **Stacked on the PHASE-1 PR.** Base is `CSDK/phase-1-quick-wins`, so the diff here is the three commits below; everything under them is reviewed in that PR and in #603. GitHub retargets this PR down the stack automatically as each base merges.

PHASE-2 of the dependency modernization plan, plus the 5.1.0 release commit. **Dev-only changes: no source edits, no change to what `dist/` emits, and `engines.node` stays `">=18"` for consumers.** Three commits, each reviewable on its own.

## `20ff94e8` — Modernize the dev toolchain and isolate the docs site

**ESLint 7 → 10 on flat config.** `.eslintrc.js` is replaced by `eslint.config.js`, and the split `@typescript-eslint/{parser,eslint-plugin}@5` by the `typescript-eslint@8` meta-package. The config is a deliberate 1:1 port of the old rule set, which is why it lands with zero source changes. Three decisions are worth confirming rather than skimming:

- ESLint core's `recommended` is **not** enabled, because the old config never enabled it either (`@typescript-eslint/recommended` does not imply it). Turning it on surfaces 61 pre-existing style findings in `src/` that belong in their own change.
- `@typescript-eslint/no-unused-expressions` is off. It is new in v8's recommended set, and 116 of its 117 hits are chai assertions.
- `no-unused-vars` keeps `caughtErrors: 'none'`; v8 flipped that default.

**husky 4 → 9 and lint-staged 10 → 17.** This also repairs a genuinely broken mixed state: `package.json` carried a v4 `husky.hooks` block and `prepare: "husky-run install"` while a v5-style `.husky/pre-commit` also existed — one that ran the full test suite including e2e, plus a stray `pre-commit` line that would have failed as command-not-found. The hook now runs `lint-staged` only, and the obsolete `git add` step is dropped, since lint-staged has auto-staged since v11.

**Docs site isolation.** `site/` moves to its own `package.json` with its own lockfile rather than an npm workspace — a workspace shares the root lockfile, so `vocs` would still have shown up in the SDK's audit. This removes `vocs` and eight override hacks from the root tree, and with them the `@hono/node-server`, `react-router` and `nuqs` advisories. The `docs:*` scripts install `site/` on demand.

**Also:** prettier 1 → 3, the webpack stack (webpack 5.108, webpack-cli 7, copy-webpack-plugin 14, webpack-bundle-analyzer 5, webpack-node-externals 3, ts-loader 9.6), sinon 15 → 22 with `@types/sinon` and `@types/mocha` aligned, and a `serialize-javascript ^7.0.7` override.

That override is the one deferred out of PHASE-1. `serialize-javascript@7` reads global `crypto` at load and so needs Node ≥19 — safe here because `webpack-cli@7` and friends require Node ≥20.9 anyway, so the lint and build CI steps are now gated to the 22/24 legs, and mocha only requires it from its `--parallel` worker pool, which this repo never uses. The 18.x leg keeps install, audit and the unit suite.

The lockfile is regenerated because npm could not resolve incrementally against the stale `@typescript-eslint@5` entries. **The production dependency closure is unchanged: all 45 packages resolve to the same versions.**

Full `npm audit` drops from 14 advisories to 7, with all HIGH cleared. `npm audit --omit=dev` remains 0.

## `60d948ce` — Apply the Prettier 3 reformat

Mechanical output of the prettier 1 → 3 upgrade, kept in its own commit so neither change has to be reviewed through the other. The blast radius is small: prettier only ever runs on `*.{js,css,md}` via lint-staged, never on `src/*.ts` (those are `eslint --fix` only), and `.prettierrc` already pins the options v2 flipped by default (`trailingComma: 'none'`, `arrowParens: 'avoid'`). What is left is pure v1 → v3 algorithm drift: a space after an anonymous `function`, and assignments that now break after `=` rather than before the member access. Adds `.prettierignore` for `dist/`, `docs/`, `node_modules/` and `site/` — the docs site is its own package now and formats on its own terms.

## `58a46965` — Release 5.1.0

Version bump and the changelog entry covering PHASE-1 and PHASE-2 together. The entry leads with what matters to consumers: no runtime or API change, `dist/` behaves as in 5.0.13, `engines.node` unchanged — the visible effects are a smaller production dependency tree and refreshed transitive versions.

One backfill worth flagging: the entry for **5.0.13** now also documents the `files` whitelist introduced in `e3e4248c`, which that release's own entry omitted. It is a user-visible packaging change — the `site/` sources, `.github/`, `.husky/` and the build configs no longer ship — and without the bullet, anyone moving 5.0.12 → 5.1.0 would have no record of it. #603 may want the same bullet.

## Verification

On Node 22.22.2 against the release commit: `lint:ci` clean at `--max-warnings=0`, `npm run knip` exit 0, 170 unit tests passing, `npm audit --omit=dev` 0 vulnerabilities, `npm run build` green, `prettier --check` clean, and `npm pack --dry-run` produces the expected whitelisted tarball. The commits themselves were verified across the 18/22/24 legs as they landed.

`npm run test:browser:unit` remains **known red and pre-existing**, unchanged by this PR — see the PHASE-1 PR for the diagnosis. PHASE-3 removes karma outright.


davidatwhiletrue (approved):
LGTM. I would summarize the CHANGELOG though.

PR #610, opened 2026-08-11
Phase 3: Vitest, TypeScript 6 and dependency migrations (5.2.0)

Comp0te:
First PR in a stack. It carries the phase-3 toolchain and dependency migrations that everything above it builds on, and releases them as `5.2.0`.

`dist/` behaves as in `5.1.0` for the vast majority of consumers: nothing was removed or renamed, and `engines.node` stays `">=18"`. Three changes do reach consumers and are called out below.

## Test tooling

- mocha, chai, karma, nyc and ts-node are replaced by Vitest. The same suite runs under Node and, via Playwright, in a real Chromium (`test:node:unit` / `test:browser:unit`); the e2e runner moved to Vitest as well.
- `emitDecoratorMetadata` is off for the Vitest transform. oxc emits bare identifiers where tsc guards them with `typeof`, so a class used as a parameter type before its own declaration throws a TDZ error at import time.
- A `no-restricted-globals` rule over `src/**` (minus `src/tests/**`) catches test globals leaking into shipped code. `vitest-globals.d.ts` is ambient and applies to the whole program, so a stray `describe` in `src` used to type-check in the editor and pass lint, failing only at `npm run build`.

## TypeScript 6

- TypeScript 4.9 → 6.0.
- Six members had no declared return or parameter type, so the compiler printed its own inference into the emitted declarations: `StoredTarget.toBytes` and `splitAt` gained the generic `Uint8Array<ArrayBufferLike>` form, which does not compile below TypeScript 5.7 with `skipLibCheck: false`, and `parseKey`, `Hash.fromBuffer` and the two secp256k1 encoder helpers named `Buffer` outright. **Consumer-visible:** TypeScript 6 no longer emits `/// <reference types="node" />`, so any `Buffer` left in a signature would have made the published `.d.ts` require `@types/node` for the first time. All six are annotated; `Buffer` is a `Uint8Array`, so every call site is unaffected. `dist/**/*.d.ts` now type-checks with `types: []` and `skipLibCheck: false`.

## Dependencies

- `asn1.js` and `bn.js` are gone, replaced by `@peculiar/asn1-ecc`. The emitted DER and PEM are byte-identical, so existing key files keep parsing. `asn1.js` had been unmaintained since 2020 and only worked here because the package forced `bn.js@5` on it through an override. secp256k1 import now also rejects keys that are structurally valid but not secp256k1 — a P-256 PEM, a wrong-length private key or an unexpected `ECPrivateKey` version used to be accepted silently.
- `eventsource` 2 → 3. **Consumer-visible:** errors now arrive as an exported `SseError` carrying the HTTP status code, and `SseClient.start()` accepts an optional error callback. Without a callback it no longer throws — it logs and lets the stream reconnect. The old `throw` ran inside `eventsource`'s own fetch promise chain, where an HTTP failure was swallowed whole (the stream died silently) and a transport failure escaped as an unhandled rejection that terminated the host process, cancelling the reconnect it was about to schedule.

## Latent defects surfaced by the migration

These predate this branch and only became reachable once the bundles stop being ES5 (the test transform today, `target: es2020` in v6):

- `TransferHash` and `TransactionHash` called their base constructor conditionally, leaving the object half-built on some paths.
- Two import cycles (`Key` ↔ `Account`, `Transform` ↔ `TransformRaw`) threw at import time under native ES modules. `PrefixName` and `NamedKeyKind` moved into their own modules and are re-exported from their previous homes, so every import path still resolves.
- `InfoGetStatusResult.available_block_range` was declared with a `constructor:` thunk returning an object of decorators rather than a class, which typedjson rejects: the ES5 output drops the field silently, native ES classes abandon the whole document.
- `Hash.equals()` read the other object's private field, bypassing subclass overrides, so `deployHash.equals(transaction.hash)` and its reverse disagreed for the same pair.
- `SseError` lost its prototype chain in the ES5 bundle, making `instanceof` false for a genuine `SseError`.
- The PEM writer emitted a blank line before the footer when the base64 body was an exact multiple of 64 characters, producing a file OpenSSL and Node's `crypto.createPrivateKey` both reject.

## Wire-format compatibility

Nothing here was supposed to change what the SDK reads or writes, and nothing checked it. A recorded compatibility suite captures what `casper-js-sdk@5.1.0` — installed from the registry, not built here — produced over this repo's fixture corpus, and replays every case against the working tree: 22 JSON round-trip digests, 9 transaction and deploy byte encodings, 49 `Key` vectors sampled from every prefix, 6 `URef` vectors, the `PrefixName` enum, PEM/DER key material and signatures for both curves, and four secp256k1 DER vectors pushed back through 5.1.0's own codec. Byte equality is what makes it bidirectional: a 5.1.0 artifact still parses here, and anything written here is still readable by 5.1.0. The two deliberate divergences (`Hash#equals`, `InfoGetStatusResult`) are documented in the suite.

## CI

- A Node 18 consumer job installs the built tarball and exercises it. Its smoke script used to verify each signature against the hash it had just been made over, which holds for any payload — a reordered field or wrong length prefix would have shipped green. The builder's timestamp and TTL are pinned and both algorithms sign deterministically, so the hash, signature and full serialized bytes are asserted as exact hex.
- Every action in the CI and publish workflows is pinned by SHA, at the commit each existing major tag resolves to rather than bumped to a newer major.
- The web build's `resolve.fallback` block is gone — every entry was proven unreachable, so reaching for a Node builtin now fails the build instead of silently bundling a polyfill.


src/rpc/response.ts
```diff
@@ -762,14 +774,16 @@ export class InfoGetStatusResult {
   })
   latestSwitchBlockHash: Hash;
 
+  // `constructor` must name a real class here. A thunk returning an object of
```
davidatwhiletrue: remove this comment before merging.

src/sse/client.ts
```diff
@@ -11,6 +11,45 @@ import { EventParser } from './event_parser';
...
+    super(
+      message ??
+        (code
+          ? `Event stream failed with status ${code}`
+          : 'Event stream failed')
+    );
+    this.name = 'SseError';
+    this.code = code;
+
+    // `target: es5` downlevels `extends Error` to `_super.call(this, …) || this`;
```
davidatwhiletrue: remove this comment before merging.

davidatwhiletrue (approved):
LGTM. Consider cleaning some comments added by agents with explanations that won't help in the future... (e.g. src/rpc/response.ts Ln 777)

Comp0te:
Thanks — good catch on the comments.

I swept every comment this PR adds under `src/**` (tests aside) and trimmed the pass. The fix is pushed to the **top of the stack**, [#614](https://github.com/casper-ecosystem/casper-js-sdk/pull/614) (`CSDK/phase-4.5-test-coverage`), commit `0de0c66f`, rather than rebasing four branches to place it here — the files it touches are unchanged by #611/#613/#614, so the end state of the stack is identical either way. Say the word if you would rather see it on this branch and I will rebase the stack.

What changed:

**Removed outright** — `InfoGetStatusResult` (the one you pointed at, `src/rpc/response.ts` Ln 777), the `fromBuffer` docstring in `key/Hash.ts`, and three notes in `keypair/` explaining why `Buffer` was widened out of a signature. That last rule is not lost: `check:package` enforces it in CI, and `tsconfig.dts-consumer.json` explains it where it is enforced.

**Trimmed to the constraint that still binds** — the rest. The `super`-must-be-unconditional notes, the two "this module must stay a leaf" headers, `Hash#equals`, the PEM chunking, and the SSE "never throw from here" rule now state the rule instead of narrating what `target: es5` used to hide or what TypeScript 6 stopped emitting. The `asn1.js` obituary in the secp256k1 header is gone; the ASN.1 schema and the pointer to the pinned vectors stayed.

Lint and the unit suite (719 tests) pass; no emitted output changes.

PR #611, opened 2026-08-12
Phase 3.5: ESLint hardening and TypeScript strict

Comp0te:
Second PR in the stack, on top of #610. It turns on the correctness checks phase 3 left off — ESLint's core recommended set plus the type-aware rules, and TypeScript `strict` — and fixes what they found. No public API change, no version bump.

## TypeScript strict

- `tsconfig.base.json` set seven strict-family flags individually and left four unmeasured. Three are free: `strictFunctionTypes` costs two type arguments, `strictBindCallApply` costs nothing, `useUnknownInCatchVariables` costs thirteen catch sites. Only `strictPropertyInitialization` is a real project (389 errors across the typedjson-decorated classes) and stays pinned off with a note. Setting `strict: true` rather than listing flags means future TypeScript releases add their new checks here automatically.
- `toError` gives the thirteen catch sites one coercion rule. It is deliberately not re-exported from `src/utils/index.ts`, which `src/index.ts` re-exports wholesale — listing it there would make an internal helper public API.
- `lib` moves to es2022 so `Error.cause` type-checks. `target` stays es5, so this is a type-checking change only: building the untouched tree with only the compiler flags flipped emits a byte-identical bundle.

## The byte-to-enum boundary

This SDK decodes discriminated unions from bytes — a tag byte picks a CLType, a TransactionEntryPoint, a `Key` variant, a KeyAlgorithm. Two failure modes followed and neither was caught: adding a union variant without its `case` compiled fine and fell silently to `default`, and a `number` read off the wire was compared against an enum member 47 times without sharing its type. A renumbered enum or a mistyped literal would have selected the wrong branch and produced a well-formed but wrong transaction — a valid signature over the wrong intent, not an exception.

Each decoder now narrows the raw byte once, where it enters the domain, instead of casting at each comparison. The `undefined` branch throws exactly what the old `default:` threw, so no input changes which branch it takes; where the switch became exhaustive the unreachable `default:` is gone. Adding a member to `TypeID` now fails the lint gate in three files, two of which this change never touched.

## Error handling

- Twelve rethrows discarded the error they were wrapping, turning a debuggable network or parse failure into a bare "Failed to parse" on the consumer's side. Each now carries `{ cause }`; every message string is untouched, since consumers match on those. On a pre-ES2022 runtime the options argument is ignored and the cause is simply absent, which is additive.
- **Behavior change on malformed input:** `CLValueMap.fromBytes` swallowed any failure parsing an entry and continued. A key that parsed before its value threw had already advanced the remainder, so callers got a map quietly missing entries and a byte offset pointing into the middle of one — every subsequent decode misaligned. It now throws with the cause attached.
- `putTransaction` rejected with a bare string when handed a non-legacy transaction on a 1.5 network. Same text, now an `Error`.
- The eight `return promise` sites now await. None sat inside a `try`, so none was the bug that rule usually finds — they are a tripwire for the next one that is.

## ESLint

- `js.configs.recommended` was off in phase 2 because that phase forbade `src/` edits, so the baseline almost every JavaScript project runs had never applied here. Added on top: `no-shadow`, which matters where a decoder reuses `bytes` or `serializer` in a nested scope, and `no-param-reassign` — no `Uint8Array` was being mutated; all six reassignments were strings and numbers, now locals. `no-shadow` is relaxed for exactly two files, where enum members are named after classes those files import; renaming a member would break a published enum.
- Twenty-two zero-cost rules catch a dropped promise, an eval, a switch fallthrough, a promise passed where a boolean is expected. All but one already reported zero — the exception was `sleep`, whose promise executor returned `setTimeout`'s handle.
- `reportUnusedDisableDirectives` is an error, so a suppression cannot outlive the problem it hid.

## Deprecated APIs in the signing path

The fifteen `no-deprecated` findings sat in the code that signs. Four were noble import specifiers that moved upstream (`sha512`, `sha256` ×3, `blake2b`) — same symbols, same implementations, only the path moved; noble itself is not upgraded here. The nine `slice` sites are all on a Node `Buffer`, whose `slice` is deprecated precisely because it already returns a view rather than a copy, so `subarray` is an exact rename and no caller's buffer is newly aliased. `Hash.ts` and `BidAddr.ts` take a real `Uint8Array`, where the copy is load-bearing; neither was flagged and neither was touched.

`signing-vectors.test.ts` records the ed25519 and secp256k1 signatures for fixed keys. It was written and run against the pre-change tree first, so it asserts yesterday's bytes.


davidatwhiletrue (approved):
LGTM

PR #613, opened 2026-08-13
Phase 4: supply-chain hardening and release gates

Comp0te:
Third PR in the stack, on top of #611. It closes the supply-chain and release gaps this repo carried into 5.x: install scripts running unchecked, no way to tell what shipped in a release, and a public API surface protected only by review convention. No source behaviour change, no version bump — the diff is workflows, config and two scripts.

## Install scripts and the dependency window

A malicious `postinstall` is the npm attack that needs no cooperation from this repo's code. `.npmrc` sets `ignore-scripts=true`, and the allowlist that replaces it is empty — `allow-scripts auto` found exactly one package in the whole tree declaring a lifecycle script, the tripwire itself. `@lavamoat/preinstall-always-fail` fails the next install loudly if the flag is ever removed, which it demonstrated by failing the install that added it.

The flag is blunt, and part of this PR is paying that bill. npm no longer runs `pre`/`post` scripts, so `prebuild` silently stopped cleaning and linting — `build` chains those itself now. `prepublishOnly` no longer runs on publish at all, so the tests, package-shape checks and API tripwire it carried are explicit steps in the release workflow, where they also fail the tag before anything is uploaded. `prepare` no longer installs the git hooks, so contributors run `npm run setup`. `site/` is untouched: npm treats it as its own project and does not read the root `.npmrc`, so the vocs tree still needs its own handling, noted at the call site.

The hook itself is then deleted rather than left to read as if a hand-rolled publish were gated — confirmed on npm 10.9.7 and 11.19.0 that `prepublishOnly` never fires on any path this repo publishes from.

Dependabot raises routine updates in one weekly window, with dev bumps grouped into a single PR and actions into another. The signing-path packages stay out of that group: a `@noble` or `@peculiar` change reaches key material and the DER codec, so it gets its own PR and its own review rather than arriving inside a batch. `versioning-strategy` is `increase-if-necessary` because the caret ranges in `package.json` are a contract with consumers, and the default would narrow them on every minor bump.

## What gets published, and what is in it

`npm publish` gains `--provenance`, so npmjs.com can attest which repo, workflow and commit built the tarball — the job already held the OIDC permission. `npm audit signatures` runs beside the advisory audit, covering what advisories cannot: a package whose tarball no longer matches what the registry signed.

Each release now carries a CycloneDX SBOM of what a consumer actually resolves. The reason is a repeat of `inflight@1.0.6`: no GHSA existed for it, so `npm audit` reported zero while it shipped for months, and answering "which of our releases contained it?" meant rebuilding old dependency trees by hand.

Packaging is checked rather than assumed — publint on the manifest, attw on whether the declarations resolve the way the runtime entries do, and `tsconfig.dts-consumer.json` type-checking `dist/index.d.ts` with `types: []`, so a Node global re-entering a public signature fails here rather than in a consumer without `@types/node`. attw is red on one rule and it is ignored deliberately: no named export is statically detectable inside the webpack bundle, so a Node ESM consumer's named import type-checks and then crashes at runtime. Leaving the rule on would mask every other regression the step exists to catch.

## Gates that fail instead of report

- **Public API.** api-extractor rolls the built declarations into `etc/casper-js-sdk.api.md` and CI compares the emitted surface against it, so a change to the public API has to arrive as a reviewable diff of the report in the same PR. It reads `dist/*.d.ts` rather than `src`, so it sees what a consumer's compiler sees. `ae-forgotten-export` is silenced rather than fatal, because the existing surface already leaks non-exported types.
- **Bundle size.** webpack has printed a size warning on every build for as long as `lib.cjs.js` has been over 244 KB, and a warning has never failed anything. Slow growth is not the risk; the single import that drags a whole library in is, and v6 walks straight into it. `lib.web.js` is the one that matters — it ships to browser wallets, on someone else's connection.
- **knip.** The CI gate was limited to dependency findings while the dead-code inventory was open. It is closed, so the narrow scope now just lets dead code re-accrete. `src/index.ts!` is declared rather than inferred: everything reachable from it is the published surface, and until now that carve-out existed only because the webpack plugin happened to derive the same entry. Both runs verified against a mutation probe rather than a clean report.

## Static analysis

Semgrep and CodeQL, with the overlap deliberate. Beyond the registry rulesets, `.semgrep/` encodes the invariants no generic ruleset can know here — the ones written down only in `CLAUDE.md`. `reflect-metadata` is gone, a `@jsonMember` without an explicit type silently needs it back, `super()` must be unconditional under native ES classes, and a Node builtin in shared code fails the web build. Every rule was checked twice: against a fixture of deliberate violations to prove it fires, and against the tree to prove it is quiet.

Triaging the registry rulesets' first real run found one defect worth calling out: the publish job interpolated `github.ref_name` straight into a `run:` line, on the one job holding `id-token: write`. The tag name now reaches the shell through `env:`. Exploiting it needed push access, so this is depth rather than a hole. Six findings were false and are suppressed by rule id rather than by excluding the files, so a real finding in either still lands.

## Workflow hygiene

Every workflow declares a top-level `permissions:` floor and elevates per job; a Semgrep rule now holds them to it, since a workflow that omits the key gets whatever the repository default token grants. Remaining unpinned actions are pinned by SHA. `dependency-review` blocks a PR that introduces a dependency with a known moderate-or-worse advisory.

The disabled nightly workflow is deleted rather than kept on life support: a `.disabled` suffix hides the file from GitHub and from Dependabot alike, so the SHAs pinned in it would have rotted silently. Nothing in it was salvageable — Node 16/18, a pre-2.0 `casper-node-launcher-js` carrying a HIGH advisory, an e2e script matching zero files, and an nctl faucet private key inline in `env`.

`SECURITY.md` gives vulnerability reports somewhere private to go. There was no policy, so a researcher who found a flaw in key generation or signing had one visible option: open a public issue on a repo whose consumers hold funds.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


davidatwhiletrue (approved):
LGTM

PR #614, opened 2026-08-17
Phase 4.5: test coverage, and the defects it found

Comp0te:
Fourth PR in the stack, on top of #613. The three phases below it changed the toolchain, the lint and type gates, and the supply chain, each on the claim that nothing about the SDK's behaviour moved. Very little actually checked that claim: the suite tested 34 files and 371 cases, almost all of it on the transaction path.

This PR closes that gap and reports what it found. Unlike the phases below it, this one does change behaviour — seventeen defects, all of them pre-existing, none of them visible to any test that existed before.

## What the sweep found

The bias throughout is characterization over unit assertions: parse a recorded node payload, write it back, require it to reproduce byte for byte. A `fromJSON`-only test leaves the `toJSON` direction free to break silently, which is how most of the following got in.

**Wire format**, where the SDK and the node disagreed:

- `queryGlobalStateByBlockHash`, `ByBlockHeight` and `ByStateHash` passed an object literal where a `ParamQueryGlobalStateID` was declared. typedjson refused it and dropped the member, so all three queried latest state and silently ignored the block or state root they were given.
- `PublicKey.accountHash()` carried the prefix `account-hash` without its separator, so every request param built from one was malformed.
- `StateGetEntityResult` knew only the 1.x `LegacyAccount` spelling; a 2.x `Account` payload parsed to an entity with nothing set, and no error.
- `SeigniorageAllocation.fromJSON` picked its reader by key name, and 2.x nests `delegator_kind` under the same `Delegator` key 1.x used for a flat public key — so 2.x era data lost the delegator.
- `DelegationKind` had no serializers and emitted raw object dumps, destroying the key material in every `EraInfo`, `EraEnd` and `AuctionState` document the SDK wrote.
- `EntryPointAccess.toJSON` wrote `groups`; the node writes `Groups`.
- `StepEvent` declared `step` against the node's `Step`, so the payload was always `undefined`.
- `InfoGetStatusResult` emitted no `available_block_range`.
- `BalanceHoldAddr.fromJSON` and `BlockGlobalAddr.fromString` rejected the prefixed form their own writers emit.

**Byte decoding**, where a reader trusted a computed offset instead of the previous read's remainder:

- `BalanceHoldAddr` built a `DataView` without the view's own `byteOffset`, so reached through `Key.fromBytes` — where the input is a subarray — it read a block time out of unrelated bytes.
- `EntryPointAddr`, `MessageAddr` and `NamedKeyAddr` located their second field in the original buffer, landing one byte early or on the entity address itself.
- `Key.fromBytes` rebuilt a `TransferHash` from hex rather than bytes, dropping the `transfer-` prefix.
- `toBytesU512` took its bound by masking `MaxUint256`, and a mask cannot widen a 256-bit value, so the whole upper half of the legal U512 range threw `value out-of-bounds`.
- `parseU32` returned a negative number for any value with the high bit set.

**Control flow:**

- `waitForTransaction` raised its timeout inside a `setTimeout` callback, where a throw cannot reject the surrounding promise. It escaped as an uncaught exception while the promise stayed pending and polling ran on indefinitely.
- The four SSE event `fromJSON` parsers returned an `Error` that `parseAs*Event` then handed straight back as the parsed event, so the declared return type was a lie. They throw now, and the API report records the narrowed signatures.

## For consumers

Three of these change how existing code runs, and the changelog marks them: the SSE event parsers throw where they returned an `Error`, `waitForTransaction` rejects where it left an uncaught exception behind an endless poll, and `CLValueMap.fromBytes` throws on an entry it cannot parse rather than returning a map quietly missing entries. The rest are fixes to values that were already wrong, so recorded output may differ.

Three cases also diverge from what the 5.1.0 compatibility golden recorded. Rather than re-record the golden, those cases are pinned to the digest of the corrected form and assert they no longer match what 5.1.0 wrote — accidental drift still fails, while the intended divergence stays declared in the test.

## The e2e suite

`run_e2e_locally.sh` pointed at a hand-rolled network and the suite it drove was lost; `vitest.e2e.config.mts` carried `passWithNoTests` to keep the empty run green, which is the same silence the karma suite hid behind for eighteen months. The flag is gone and an empty run fails there as it does in the unit config.

The replacement covers the read path, SSE and transaction submission against the `makesoftware/casper-nctl` image. `e2e/run.sh` brings the container up, extracts the faucet key and runs vitest against it, so a contributor needs Docker and nothing else. Every suite spends the same faucet account, so files run serially rather than racing its nonce.

`e2e-smoke` gives each change a real-node signal in about five minutes — read path plus one native transfer. The full tag × Node matrix runs weekly instead, since pinned image tags cannot drift between two PRs; what a cron does catch is the live networks moving past the images we pin, which is its own weekly job and two HTTP calls.

## Keeping it

Coverage thresholds in `vitest.config.mts` are the measured baseline rounded down. They are not an aspiration — they exist to fail the build when a change *removes* coverage. The `assert-test-count.js` floors move to 69 files and 711 cases: vitest errors on a total-zero run by itself, but cannot notice a partial loss, and thirty files pass exactly as green as seventy.


src/utils/casper-network.ts
```diff
@@ -38,6 +38,11 @@ function toRpcErrorCode(code: number): ErrorCode | undefined {
     : undefined;
 }
 
+/**
```
davidatwhiletrue, line 41: I'd remove this comment. IDEs show it as CasperNetwork description.

src/types/Bid.ts
```diff
@@ -145,9 +145,17 @@ export class DelegationKind {
   /**
    * A delegation bid made using a public key.
    */
+  // Explicit serializers: left to typedjson the key reserializes to
```
davidatwhiletrue, line 148: remove.

src/types/ByteConverters.ts
```diff
@@ -31,47 +31,47 @@ const describeValue = (value: BigNumberish): string => {
...
-): Uint8Array => {
-  const val = BigNumber.from(value);
+export const toBytesNumber =
+  (bitSize: number, signed: boolean) =>
+  (value: BigNumberish): Uint8Array => {
+    const val = BigNumber.from(value);
 
-  // Calculate the maximum allowed unsigned value for the given bit size
-  const maxUintValue = MaxUint256.mask(bitSize);
+    // Shifted, not masked off `MaxUint256`: a mask cannot widen a 256-bit value,
```
davidatwhiletrue, line 39: remove

src/types/ByteConverters.ts
```diff
@@ -177,7 +177,11 @@ export function parseU32(bytes: Uint8Array): number {
     throw new Error('Invalid byte array for u32 parsing');
   }
 
-  return bytes[0] | (bytes[1] << 8) | (bytes[2] << 16) | (bytes[3] << 24);
+  // `>>> 0` because `<< 24` produces a *signed* int32: without it every u32
```
davidatwhiletrue, line 180: remove

davidatwhiletrue (commented):
@Comp0te , I see in the changes many comments that we probably should remove. They are clarifications on why the changes are made. But, IMO they shouldn't remain after merge...

src/types/Bid.ts
Comp0te: Removed in 2782eac4.

src/types/ByteConverters.ts
Comp0te: Removed in 2782eac4.

src/types/ByteConverters.ts
Comp0te: Removed in 2782eac4.

src/utils/casper-network.ts
Comp0te: Removed in 2782eac4, along with the same class of comment across the rest of the branch — the ones that narrate what the old code got wrong. What survived is trimmed to the constraint, and the public API docs are untouched.

